PR #4, opened 2024-10-15
chore(playground): set charset in content-type header

fvsch:
The emoji in the example HTML response may not render correctly in the playground's iframe (or when running the code locally) without a charset declared in the HTML or in the `Content-Type` header.

Screenshot: a web page in a browser with the UTF-8-encoded text being rendered as latin1 characters


pi0 (approved):
thanks!🙏

PR #6, opened 2024-10-15
docs: changed to use cache of most recent images

EdamAme-x:
Since this is a minor documentation fix, I determined that an issue was unnecessary.


EdamAme-x:
Current (cache is 404)
image

Changed
image

pi0:
Thank you for the PR. It is an auto generated section and now cache is invalidated.

PR #16, opened 2024-12-01
feat(node): add `NodeFastResponse.bytes()`

bergold:
The NodeFastResponse class is missing the `bytes()` method of the Response interface. [Link to MDN](https://developer.mozilla.org/en-US/docs/Web/API/Response/bytes)


pi0 (approved):
Thanks ❤️

PR #38, opened 2025-03-12
feat: support `tls` and `protocol`

### LINKED ISSUE #19: SSL / HTTPS support
### Describe the feature

As per title 😄.

### Additional information

- [ ] Would you be willing to help implement this feature?

OskarLebuda:
Resolves #19

### DIFF
--- a/docs/1.guide/5.options.md
+++ b/docs/1.guide/5.options.md
@@ -58,7 +58,9 @@ Enabling this option allows multiple processes to bind to the same port, which i
 
 ### `tls`
 
-TLS server options. When **both `tls.cert` and `tls.key`** provided, the server will use HTTPS instead of HTTP.
+TLS server options.
+
+When **both `cert` and `key`** are provided, the server will use HTTPS instead of HTTP.
 
 **Example:**
 
@@ -73,8 +75,8 @@ serve({
 
 **Options:**
 
-- `cert`: Path or inline content for the certificate (PEM format).
-- `key`: Path or inline content for the private key (PEM format).
+- `cert`: Path or inline content for the certificate in PEM format.
+- `key`: Path or inline content for the private key in PEM format.
 - `passphrase`: Passphrase for the private key (optional).
 
 > [!TIP]
--- a/docs/1.guide/5.options.md
+++ b/docs/1.guide/5.options.md
@@ -90,6 +90,7 @@ serve({
 ```
 
 > [!IMPORTANT]
+>
 > - Always keep your SSL private keys secure and never commit them to version control
 > - Use environment variables or secure secret management for production deployments
 > - Consider using automatic certificate management (e.g., Let's Encrypt) for production
--- a/docs/1.guide/5.options.md
+++ b/docs/1.guide/5.options.md
@@ -56,38 +56,29 @@ Enabling this option allows multiple processes to bind to the same port, which i
 > [!NOTE]
 > Despite Node.js built-in behavior that has `exclusive` flag enabled by default, srvx uses non-exclusive mode for consistency.
 
-### `https`
+### `tls`
 
-HTTPS configuration options. When provided, the server will use HTTPS instead of HTTP.
+HTTPS configuration options. When provided, the server will use HTTPS instead of HTTP **if both `tls.cert` and `tls.key` are set**.
 
 **Example:**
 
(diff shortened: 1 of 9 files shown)


docs/1.guide/5.options.md
```diff
@@ -56,6 +56,45 @@ Enabling this option allows multiple processes to bind to the same port, which i
...
+
+```js
+import { serve } from "srvx";
+import { readFileSync } from "node:fs";
+
+serve({
+  // Using file paths
+  https: {
+    key: readFileSync("./server.key"),
+    cert: readFileSync("./server.crt"),
```
pi0, line 73: Since adapter are conditionally imported, we can support reading from a file path instead of making users to read it themselves.

src/adapters/node.ts
```diff
@@ -68,7 +72,13 @@ class NodeServer implements Server {
       ...this.options.node,
     };
 
-    const server = NodeHttp.createServer(this.serveOptions, handler);
+    // Create HTTPS server if HTTPS options are provided, otherwise create HTTP server
+    const server = this.isHttps
+      ? NodeHttps.createServer(
+          { ...this.options.https, ...this.serveOptions },
```
pi0, line 78: We can refactor this under `serveOptions`

src/adapters/node.ts
```diff
@@ -91,9 +101,10 @@ class NodeServer implements Server {
     if (!addr) {
       return;
     }
+    // Use https:// in URL if HTTPS is enabled
```
pi0, line 104: i think your code is already readable in L107 :)

docs/1.guide/5.options.md
OskarLebuda: I see. Do you think we should add additional options like `keyPath`, `certPath`, etc., or should we assume that it will always be a file path?

src/adapters/node.ts
OskarLebuda: removed

src/adapters/node.ts
OskarLebuda: fixed

docs/1.guide/5.options.md
pi0: i would use simpler `key`/`cert` as path/url and then something like `inlineKey` for inline versions

OskarLebuda: fixed! :)

src/types.ts
```diff
@@ -65,10 +66,53 @@ export interface ServerOptions {
    */
   reusePort?: boolean;
 
+  /**
+   * HTTPS configuration options.
+   * If provided, the server will use HTTPS instead of HTTP.
+   * You need to provide at minimum the key and cert options.
+   */
+  https?: {
```
zsilbi, line 74: I would prefer to have a boolean flag like `enabled` here to easily disable `https` in development env.

If this is set to `true` an error should be thrown instead of falling back to `http` when no certificates are provided.

OskarLebuda: I'm not sure about the `enabled` flag at all.

pi0:
Pushed some updates mainly to simplify both usage and implementation:

- `https` renamed to `tls` to as better technical term
- Reuse one util for resolving cert/key (both Deno and Bun have native node fs compatibility)
- Updated inline support to detect PEM string format to simplify options
- Removed ca and PFX support as Deno does not allow this and is not common (can be set with `node` and `bun`)

src/types.ts
pi0: I agree that we need a way to disable TLS directly. even in playground it is needed!

@OskarLebuda, What alternatives do you suggest?

pi0: We talked to introduce a new `type: "https" | "https"` option that can be extendable to more types like `http2`

src/_utils.ts
```diff
@@ -14,13 +18,52 @@ export function resolvePort(
...
+    throw new TypeError("TLS `cert` and `key` must be provided together.");
+  }
+  return {
+    cert,
+    key,
+    passphrase: opts.tls.passphrase,
+  };
+}
+
+function resolveCertorKey(value?: unknown): undefined | string {
```
zsilbi, line 56:
```suggestion
function resolveCertOrKey(value?: unknown): undefined | string {
```

PR #58, opened 2025-04-23
feat(node): support http2

OskarLebuda:
Add `http/2` support for node

Performance results:
| Before | After |
|--------|--------|
| image | image |


src/adapters/http2.ts
pi0, line 1: This can be merged with node.ts

OskarLebuda: merged!

src/types.ts
```diff
@@ -144,6 +149,16 @@ export interface ServerOptions {
   };
 }
 
+export type NodeHttpHandlerFunc = (
```
zsilbi, line 152: This is a duplicate of the `NodeHttpHandler` type from [line 319](https://github.com/OskarLebuda/srvx/blob/6491d322590485fea5fbd64b7058ccdf09b51f0e/src/types.ts#L319) of this file.

src/adapters/node.ts
```diff
@@ -89,15 +91,30 @@ class NodeServer implements Server {
       ...this.options.node,
     };
 
-    // Create HTTPS server if HTTPS options are provided, otherwise create HTTP server
-    const server = (this.serveOptions as { cert: string }).cert
-      ? NodeHttps.createServer(
+    const hasCert = (this.serveOptions as { cert: string }).cert;
```
zsilbi, line 94:
```suggestion
    const hasCert = tls?.cert !== undefined;
```

zsilbi, line 94:
```suggestion
    const hasCert = (this.serveOptions as { cert?: string }).cert;
```

src/_node-compat/headers.ts
```diff
@@ -275,3 +293,14 @@ function _normalizeValue(
   }
   return typeof value === "string" ? value : String(value ?? "");
 }
+
+function isBannedHeader(name: string): boolean {
```
pi0, line 297: This logic is specific to http2 support only. We should seperate by exntending base classes or make new ones for http2.

src/_node-compat/send.ts
pi0, line 1: Combining logic of send at very least can cause performance overheads. It should be somehow separated too

src/adapters/node.ts
OskarLebuda: Fixed

src/_node-compat/headers.ts
pi0: (seems we still need to move this?) PS this check could be simplified to `name[0] === ":"` inlined in places needed it for all psudo headers

OskarLebuda: Yes, work is in progress :) Just pushed other changes

src/_node-compat/headers.ts
```diff
@@ -104,7 +105,12 @@ export const NodeRequestHeaders: {
     }
 
     *entries(): HeadersIterator<[string, string]> {
-      const _headers = this._node.req.headers;
+      const _headers = Object.fromEntries(
```
pi0, line 108: This has performance cost (and memory usage) to iterate in non http2 servers. We should find a way to either subclass or add a flag to skip when not using http2 (benchmarks might not show it)

test/node.test.ts
```diff
@@ -1,21 +1,81 @@
...
+    name: "http2 - allowHTTP1",
+    Response: globalThis.Response,
+    useHttp2Agent: true,
+    serveOptions: { tls, node: { http2: true, allowHTTP1: true } },
+  },
+  {
+    name: "http2, FastResponse - allowHTTP1",
+    Response: FastResponse,
+    useHttp2Agent: true,
+    serveOptions: { tls, node: { http2: true, allowHTTP1: true } },
```
pi0, line 32: Here we badically fallback to http1 handler already coveted what is point of it?

OskarLebuda: That was a test - forgot to remove :)

src/_node-compat/headers.ts
OskarLebuda: Right, fixed :)

src/_node-compat/headers.ts
```diff
@@ -279,8 +279,8 @@ function _normalizeValue(
 }
 
 function validateHeader(name: string): string {
-  if (name[0] === ":" || name.includes(":")) {
-    throw new TypeError("Invalid name");
+  if (name[0] === ":") {
```
OskarLebuda, line 282: I don't think that's correct. Tests could create a header like `x-app:test=1`, which is also invalid.

OskarLebuda: @pi0 ☝️

pi0: tests was wrongly creating it because input (request headers) was feeding in invalid `:test` header. in between checks add overhead (full string chars should be checked)

PR #63, opened 2025-04-24
feat: support `upgrade` hook (experimental)

pi0:
This PR is initial attempt to support unified websocket upgrade support.

- For Node.js, with `uprade` hook a Request proxy is created that adds upgrade details (duplex and head data)
- For deno, bun and cloudflare, if the request has `upgrade` header with `websocket` value and `upgrade` hook is defined, it will be called.

The next step is adding a unified `.upgrade(request)` support to srvx adapters, which can be auto-integrated.

(bun requires additional `bun.websocket`, we need to make a srvx plugin anyway to support crossws universally)

### DIFF
--- a/src/_node-compat/request.ts
+++ b/src/_node-compat/request.ts
@@ -1,8 +1,18 @@
 import type NodeHttp from "node:http";
+import type NodeStream from "node:stream";
 import type { ServerRequest, ServerRuntimeContext } from "../types.ts";
 import { kNodeInspect } from "./_common.ts";
 import { NodeRequestHeaders } from "./headers.ts";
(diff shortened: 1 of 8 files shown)


horvbalint:
I am not sure if it is relevant here, but I opened an issue in [h3](https://github.com/h3js/h3/issues/999) about some websocket problems I came across. If they are legit and planned to be fixed, they might be relevant here, so I am linking it just in case :)

Thanks for all your work!

pi0:
Yes actually this way we can first pass through h3 upgrades with normal request so hooks and middleware can be applied!

horvbalint:
Awesome, thanks a lot!

Issue #72, opened 2025-05-15
Broken headers when bundling with esbuild on Deno (Node.js compat)


finxol:
### Environment

srvx: 0.7.1

esbuild: 0.25.4

deno: 2.3.1 (stable, release, aarch64-apple-darwin)
v8 13.5.212.10-rusty
typescript 5.8.3

### Reproduction

https://github.com/finxol/repro-srvx-deno-esbuild

Instructions in README file

### Describe the bug

Creating a server with srvx works on Deno without bundling.

But when bundling it with esbuild first, the headers get broken.

###### Example from the reproduction

Without bundling:

```
< HTTP/1.1 200 OK
< content-type: text/plain;charset=UTF-8
< vary: Accept-Encoding
< content-length: 17
```

With bundling:

```
< HTTP/1.1 200 OK
< c: o
< t: e
< vary: Accept-Encoding
< transfer-encoding: chunked
```

Notice how content-type seems to be broken into `c` and `t`.

### Additional context

Bundling works fine with Node.

### Logs

```sh

```

pi0:
When building with esbuild you are bundling srvx with `node` condition.

finxol:
Changing the platform to "neutral" or "browser" doesn't work either.

```
await build({
    entryPoints: ["main.ts"],
    bundle: true,
    platform: "neutral",
    outdir: "out",
    format: "esm", // same bug with cjs format (default)
    external: ["node:*"]
});
```

The bundled server exits immediately:
```sh
finxol » deno run --allow-env --allow-net out/main.js
Server listening on undefined
```

finxol:
Thank you for the quick reply!

Adding esbuild option `conditions: ["deno"]` works great.

Importing `srvx/deno` also works, but that doesn't seem documented anywhere.

Anyway, great tool, and thank you for all your work on UnJS. :heart:

pi0:
Cool! There is a documentation section i explained but I am not best at it! Feel free to contribute to improve it https://srvx.h3.dev/guide/bundler

finxol:
Yes, I saw this page, but I thought I was fine since the bundling and running is all happening in the same runtime, even with the target platform being `node`.

I'll try to make a PR adding a bit more detail on that :)

PR #73, opened 2025-05-16
docs(guide/bundler): improve bundle usage explanation

finxol:
resovles #72 

Improves explanations for using srvx with a bundler.


pi0 (approved):
Thank you for the PR ❤️ 

While std-env idea is really intresting, i have removed it for now since cross-runtime bundling can cause similar issues. better to be explicit

PR #79, opened 2025-05-29
fix(node): only use `req.headers` in `FastResponse` when initialized

### LINKED ISSUE #78: Headers on Node.js FastResponse are duplicated if accessed via `res.headers`
### Environment

srvx: 0.6.0 (relevant code doesn't seem to have changed in later versions)
Node.js: 24.0.1

### Reproduction

1. Run the below snippet with `tsx ./repro.ts` / `node ./repro.js`
2. Execute `curl --dump-header - http://localhost:3002/redirect`. The `Location: ...` shows up twice in the output.
3. Comment out the `res.headers;` accessor and repeat the request. The `Location: ...` header is only observed once.

```ts
// repro.ts / repro.js
import { FastResponse, serve } from "srvx";

const server = serve({
	fetch(request) {
		if (request.url.includes("/redirect")) {
			const res = new FastResponse("", {
				status: 302,
				headers: {
					Location: "http://localhost:3002/target",
				},
			});

			// access res.headers;
			res.headers;

			return res;
		}

		return new Response("👋 Hello there!");
	},
	port: 3002,
});
```

### Describe the bug

Accessing `NodeFastResponse#get headers` makes a copy of `#init.headers` as observed in the following snippet:

https://github.com/h3js/srvx/blob/b8a61e13ab03210c109126df93ae2153df032593/src/adapters/_node/response.ts#L181-L183

The `#nodeResponse()` method then combines both `#init.headers` & `#headersObj`:

https://github.com/h3js/srvx/blob/b8a61e13ab03210c109126df93ae2153df032593/src/adapters/_node/response.ts#L81-L107

I want to create a PR for this, just not sure what direction this should go; `#headersObj` is only set if the headers are accessed, so we could only use `#headersObj` when it exists. However, in theory someone could keep a reference to the headers as set in the `headersInit` and modify those after accessing `response.headers` rendering `#headersObj` outdated. 

My use case for accessing `res.headers` is that I have a 'middleware' (not yet on 0.7.x middleware yet :)) that wraps a route handler and appends or overwrites application specific headers on the response returned by that route handler.

### Additional context

_No response_

### Logs

```sh

```

dirkdev98:
Resolves #78

Also made sure to prioritize `#headersObj` when using `res.clone()` or on internal access of `#response`. We are already on slow paths there, but gets more slow with the extra clone of the headers.

### DIFF
--- a/src/adapters/_node/response.ts
+++ b/src/adapters/_node/response.ts
@@ -78,13 +78,8 @@ export const NodeResponse: {
       const headers: NodeHttp.OutgoingHttpHeader[] = [];
 
       const headersInit = this.#init?.headers;
-      if (headersInit) {
-        const headerEntries = Array.isArray(headersInit)
-          ? headersInit
-          : headersInit.entries
-            ? (headersInit as Headers).entries()
-            : Object.entries(headersInit);
-        for (const [key, value] of headerEntries) {
+      if (this.#headersObj) {
+        for (const [key, value] of this.#headersObj) {
           if (key === "set-cookie") {
             for (const setCookie of splitSetCookieString(value)) {
               headers.push(["set-cookie", setCookie]);
@@ -93,9 +88,13 @@ export const NodeResponse: {
             headers.push([key, value]);
           }
         }
-      }
-      if (this.#headersObj) {
-        for (const [key, value] of this.#headersObj) {
+      } else if (headersInit) {
+        const headerEntries = Array.isArray(headersInit)
+          ? headersInit
+          : headersInit.entries
+            ? (headersInit as Headers).entries()
+            : Object.entries(headersInit);
+        for (const [key, value] of headerEntries) {
           if (key === "set-cookie") {
             for (const setCookie of splitSetCookieString(value)) {
               headers.push(["set-cookie", setCookie]);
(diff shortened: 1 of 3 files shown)


test/_tests.ts
```diff
@@ -136,4 +146,21 @@ export function addTests(opts: {
...
+    });
+
+    test("clone with headers", async () => {
+      const response = await fetch(url("/clone-response"), {
+        headers: {
+          "x-clone-with-headers": "true",
+        },
+      });
+      expect(response.status).toBe(200);
+      expect(response.headers.get("x-clone-with-headers")).toBe("true");
```
dirkdev98, line 163: Without this fix, this would be `null`.

src/adapters/_node/response.ts
```diff
@@ -139,6 +138,8 @@ export const NodeResponse: {
       // Free up memory
       this.#body = undefined;
       this.#init = undefined;
+      this.#headersObj = undefined;
+      this.#responseObj = undefined;
```
dirkdev98, line 142: I'm unsure about these, but they made sense to be here, like in 

https://github.com/h3js/srvx/blob/b8a61e13ab03210c109126df93ae2153df032593/src/adapters/_node/response.ts#L166-L175

PR #81, opened 2025-06-03
feat: universal `request.waitUntil`

### LINKED ISSUE #80: porting `event.waitUntil` from nitro to h3
### Describe the change

Hi, I was just looking for a bit more info on `event.waitUntil`. I noticed that it was referenced in the [Nitro docs for Edge workers](https://nitro.build/guide/cache#edge-workers) that `event.waitUntil` is used to do work after the response has been sent. I also see that in [this comment](https://github.com/nitrojs/nitro/issues/1705#issuecomment-1711497734) @pi0 mentions that the `waitUntil` implementation is platform agnostic. 

I think it would be useful to reference it in the 'methods' section of the 'Event Object' page. I'd be happy to write something about `event.waitUntil` generally or some common use cases. However, I don't know enough about how it is implemented in H3 / Nitro or how it behaves on different platforms. I'm just using it with the `node-server` Nitro preset in a docker container currently. 

References:
- https://nitro.build/guide/cache#edge-workers
- https://github.com/unjs/h3/issues/73#issuecomment-1131429799
- https://github.com/nitrojs/nitro/issues/1705#issuecomment-1711497734
- https://developer.mozilla.org/en-US/docs/Web/API/ExtendableEvent/waitUntil
- https://developers.cloudflare.com/workers/runtime-apis/context/#waituntil

### URLs

https://h3.unjs.io/guide/event

### Additional information

- [x] Would you be willing to help?

pi0:
resolves #80

### DIFF
--- a/docs/1.guide/2.handler.md
+++ b/docs/1.guide/2.handler.md
@@ -42,6 +42,32 @@ serve({
 });
 ```
 
+### `request.waitUntil?`
+
+Tell the runtime about an ongoing operation that shouldn't close until the promise resolves.
+
+```js
+import { serve } from "srvx";
+
+async function logRequest(request) {
+  await fetch("https://telemetry.example.com", {
+    method: "POST",
+    body: JSON.stringify({
+      method: request.method,
+      url: request.url,
+      ip: request.ip,
+    }),
+  });
+}
+
+serve({
+  fetch: (request) => {
+    request.waitUntil(logRequest(request));
+    return "OK";
+  },
+});
+```
+
 ### `request.runtime?.name?`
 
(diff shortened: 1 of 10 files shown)


NamesMT:
Hi @pi0, amazing stuff!,

From a quick read-through, if I understand it correctly, this would block the response until all Promises are resolved, right?
Would be great if you can add another version (`waitUntilAsync`?) that leverages Vercel/Cloudflare native `waitUntil` when possible, which doesn't block the response.

NamesMT:
@pi0 
Ah amazing!, How would it works with servers like AWS Lambda, which kill the process when a response is returned?

NamesMT:
Hmmm I guess `srvx` can't be used for AWS Lambda without some customizations, so that's out of scope?

PR #91, opened 2025-06-29
feat: experimental cli

dahlia:
Resolves <https://github.com/h3js/srvx/issues/90>.

Implements a command-line interface that allows starting HTTP servers directly from the command line, similar to `deno serve` and `bun run`.

Features:

- Auto-detect runtime (Node.js, Deno, Bun)
- Support for export default pattern with fetch handler
- Command-line options for port, host, HTTPS, TLS certificates
- TypeScript support where runtime allows

Usage:

```sh
srvx server.js srvx --port=8080 --host=localhost server.js srvx --https --cert=cert.pem --key=key.pem server.js
```


src/cli.ts
```diff
@@ -0,0 +1,118 @@
+#!/usr/bin/env node
```
pi0, line 1: We could use a shebang like [this](https://github.com/pi0/pipress/blob/bd8011823c54dec9d034a2ef93281e77c86d7fc5/server/index.mjs#L1-L2) to be more inclusive.

src/cli/parser.ts
```diff
@@ -0,0 +1,157 @@
...
+  /** Show version */
+  version?: boolean;
+  /** Runtime to use (auto-detected if not specified) */
+  runtime?: "node" | "deno" | "bun";
+}
+
+/**
+ * Parse command line arguments
+ */
+export function parseArgs(args: string[]): CLIOptions {
```
pi0, line 31: As both Deno and Bun support Node compat, we can use Node.js [util.parseArgs](https://nodejs.org/api/util.html#utilparseargsconfig)

src/cli/loader.ts
pi0, line 1: As this is an experimental feature, let's keep it simple and without external configuration support. 

Configuration should also anyway be passed from main exports object.

src/cli.ts
```diff
@@ -0,0 +1,118 @@
...
+ * srvx CLI - Universal HTTP server command line interface
+ */
+
+import { parseArgs, showHelp, validateOptions } from "./cli/parser.ts";
+import { detectRuntime, getAdapterImport } from "./cli/runtime.ts";
+import { loadUserFile } from "./cli/loader.ts";
+import packageMetadata from "../package.json" with { type: "json" };
+
+// Get package version
+function getVersion(): string {
```
pi0, line 13: For code style, can you please move internal utils below main export also we can move parser logic to same file as it is relavant.

src/cli/runtime.ts
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  // Default to Node.js
+  return "node";
+}
+
+/**
+ * Get the appropriate srvx adapter import path for the detected runtime
+ */
+export function getAdapterImport(runtime: SupportedRuntime): string {
```
pi0, line 28: We do not need this. Simply importing `srvx` in each runtime, loads serve function for it based on export contidions.

productdevbook:
[citty](https://github.com/unjs/citty) Do you plan to use it in this system?

pi0:
@productdevbook since srvx is zero dependency, probably not. We can still build a good CLI with node built-in utils.

PR #103, opened 2025-08-22
feat(adapter): new uWebSockets adapter

### LINKED ISSUE #71: Node.js uws adapter
### Describe the feature

[uWebSockets](https://github.com/uNetworking/uWebSockets.js) has a really fast HTTP server implementation for Node.js (not necessarily for websockets)

We can make an opt-in srvx/uws adapter (with optional peer dep) and similar wrappers of node, but for uws. 

We need to create similar to node adapter Request/Response/Headers stubs on top of uws req/res. ([example](https://github.com/h3js/crossws/blob/84cf6794f207bd5bc53be67bde68dd7e61e851e7/src/adapters/uws.ts#L222) in crossws)

### Additional information

- [x] Would you be willing to help implement this feature?

sandros94:
Resolves #71 

I wanted to take a shot at this, as it is a topic which I'm new but would like to understand. As such I want to warn that I've used extensively various LLMs to find docs and explain things, but final touch/edits/verifications were always human.

I tried to setup some benchmarks to distinguish from native node and uws, but failed. @pi0 I would love your help if and when you are available.

Tests run super smooth tho

### DIFF
--- a/build.config.mjs
+++ b/build.config.mjs
@@ -19,6 +19,7 @@ export default defineBuildConfig({
           "cloudflare",
           "generic",
           "service-worker",
+          "uws",
         ].map((adapter) => `src/adapters/${adapter}.ts`),
       ],
       rolldown: {
--- a/package.json
+++ b/package.json
@@ -88,7 +88,7 @@
     "vitest": "^3.2.4"
   },
   "optionalDependencies": {
-    "uWebSockets.js": "uNetworking/uWebSockets.js#v20.52.0"
+    "uWebSockets.js": "github:uNetworking/uWebSockets.js#v20.52.0"
   },
   "packageManager": "pnpm@10.12.4",
   "engines": {
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "./cloudflare": "./dist/adapters/cloudflare.mjs",
     "./generic": "./dist/adapters/generic.mjs",
     "./service-worker": "./dist/adapters/service-worker.mjs",
+    "./uws": "./dist/adapters/uws.mjs",
     "./cli": "./dist/cli.mjs",
     "./static": "./dist/static.mjs",
     "./log": "./dist/log.mjs",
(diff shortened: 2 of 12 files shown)


cyco130:
Thanks for taking a shot at this. I'm not a maintainer or a contributor here, so pardon the intrusion. But I've implemented a similar adapter before and there are a few sharp edges in uWebSockets.js that need to be addressed:

Most importantly, `uws.HttpRequest` is **stack allocated**. It must be "gutted" before the callback returns _synchronously_, otherwise it can be overwritten by other data and all hell might break loose. See the [docs](https://unetworking.github.io/uWebSockets.js/generated/interfaces/HttpRequest.html), the very first paragraph mentions it (but without enough emphasis, imho).

It's probably best not to expose it to the user as user code may run after any number of `await`s, where it's already too late. I would rather create a regular `Headers` object eagerly, on callback entry, and populate it with `HttpRequest.forEach`. Same goes for the URL and the method. They must be read and copied out of `HttpRequest` synchronously on callback entry, before entering any async path.

Alternatively, one could call the user handler, and if the return value is a promise, do the "gutting" at that point but I'd expect it to cause more complexity for little gain.

Another sharp edge is when reading the request body: The `ArrayBuffer` passed to the `res.onData` callback might be reused by uWebSockets.js if `isLast` is not `true`. So the data must be copied out immediately. So `controller.enqueue(new Uint8Array(chunk))` is the wrong approach. We have to do something like:

```js
const src = new Uint8Array(chunk);
const copy = new Uint8Array(chunk.byteLength);
copy.set(src);
controller.enqueue(copy);
```

for all but the very last chunk.

Also, `onData` must be attached synchronously, we can't wait until the body is read, or some data might be lost. Again, either call the user handler, and if it returns a promise without any attempt to read data, attach there (and pause it until needed). Another approach would be to attach it eagerly on entry and call pause until the body is needed.

I also see a few optimization opportunities like deferring the `AbortController/AbortSignal` pair creation until needed `onAborted` can just set a flag until then and if `signal` is accessed after `onAborted` fired, we can directly return `AbortSignal.abort()` without creating the controller. `AbortController` is somewhat expensive to create.

I also see some new APIs that I was not aware of, like `res.onDataV2` that could allow preallocating a buffer when reading the whole body.

I'm also interested in seeing uWebSockets.js adapter merged. Drop me a line [on Discord](https://discord.com/users/cyco130) or check my profile for other contact info if you'd consider pairing on this.

PR #137, opened 2025-10-31
feat(node, deno, bun): validate server port range

iiio2:


pi0 (approved):
Thanks for PR and research ❤️ 

Considering different behaviors between runmtimes, i guess is best we throw a conistent error

PR #158, opened 2026-01-13
fix(node): soft error when response body is already locked

huseeiin:
If the stream gets closed at some point (like if the browser is closed mid-stream) it can throw unhandled errors, which can crash NodeJS. ignoring these errors is completely fine and doesn't cause hidden consequences. its only a problem in the nodejs adapter and doesn't happen in deno/bun
i simply added a try catch to the whole `streamBody` function. the rest is just formatting differences

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling in streaming response processing to provide more robust failure recovery and prevent unexpected interruptions.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


huseeiin:
closing until sufficient reproduction is found

PR #174, opened 2026-01-29
feat: export `cliFetch` util

pi0:
## Summary by CodeRabbit

* **Bug Fixes**
  * CLI now exits with non-zero codes on failures and surfaces descriptive errors for non-OK responses.

* **Improvements**
  * CLI options renamed for consistency (dir instead of cwd/base, entry instead of url); downstream mappings updated.
  * Verbose mode provides clearer request/response debugging using provided I/O streams.
  * Response streaming preserves binary/text encoding behavior.

* **New Features**
  * Centralized CLI fetch path to standardize request construction, streaming, and error propagation.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -114,142 +114,6 @@ export async function main(mainOpts: MainOpts): Promise<void> {
   process.on("SIGTERM", () => cleanup("SIGTERM", 143));
 }
 
-export async function cliFetch(
-  options: LoadOptions & {
-    url?: string;
-    method?: string;
-    headers?: string[];
-    data?: string;
-    verbose?: boolean;
-    hostname?: string;
-  },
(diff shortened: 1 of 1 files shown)


src/cli.ts
```diff
@@ -39,7 +39,16 @@ export async function main(mainOpts: MainOpts): Promise<void> {
 
   // Fetch mode
   if (options._mode === "fetch") {
-    await handleFetch(options);
+    await cliFetch({
+      url: options._url,
+      entry: options._entry,
+      dir: options._dir,
```
OskarLebuda: maybe we should call this `cwd` instead of `dir`?

PR #177, opened 2026-02-09
fix(node): normalize pipeable bodies in `FastResponse` when cloning

kaigritun:
## Summary

When `clone()` is called on a `NodeResponse` with a Node.js-specific body (objects with `.pipe` method like Node.js streams), the body was lost because the `_response` getter converted it to a native `Response` which can't handle pipe-style bodies.

## Problem

This issue was reported in h3: https://github.com/unjs/h3/issues/1278

When using h3's `onResponse` middleware and calling `response.clone()`, responses with Node.js stream bodies don't get sent:

```ts
app.use(onResponse((response, event) => {
    response.clone(); // This breaks the response
}));

app.use((e, n) => {
    return {
        pipe(writable) {
            writable.write("test");
            writable.end();
        },
    };
});
```

## Root Cause

In `NodeResponse._response` getter:
1. It creates a native `Response` from `#body`
2. Native `Response` can't handle pipe-style bodies (they become `null` or broken)
3. `#body` is cleared after creating the native response
4. When `_toNodeResponse()` is called, it uses `#response.body` which is broken

## Fix

1. Before creating native `Response`, check if body is Node.js-specific (`.pipe` method)
2. Preserve such bodies in a separate `#nodeBody` field
3. In `_toNodeResponse()`, check `#nodeBody` first before falling back to `#response.body`

## Test

Added a test case that:
- Creates a response with a pipe-style body
- Calls `clone()` on it
- Verifies the original response body is still intact

The test only runs with `FastResponse` configurations since native `Response` can't handle pipe-style bodies by design.

## Checklist

- [x] Fix preserves Node.js pipe-style bodies when `clone()` is called
- [x] Added test case for the regression
- [x] All existing tests pass (651 passed)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of Node.js-style streaming response bodies so they are preserved and streamed correctly when cloned and returned.

* **Tests**
  * Added test coverage to verify cloning and streaming behavior with Node-style pipeable bodies.

### DIFF
--- a/src/adapters/_node/response.ts
+++ b/src/adapters/_node/response.ts
@@ -1,4 +1,4 @@
-import { PassThrough, Readable as NodeReadable } from "node:stream";
+import { PassThrough, Readable as NodeReadable, Writable as NodeWritable } from "node:stream";
 
 import { lazyInherit } from "../../_inherit.ts";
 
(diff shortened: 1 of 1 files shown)


src/adapters/_node/response.ts
```diff
@@ -34,6 +34,8 @@ export const NodeResponse: {
     #init?: ResponseInit;
     #headers?: Headers;
     #response?: globalThis.Response;
+    // Preserve Node.js-specific body (e.g. objects with .pipe) that can't be passed to native Response
+    #nodeBody?: NodeReadable;
```
pi0: Pipeable body, won't be actually cloned this way and shared with `res.clone()` or `res` (whoever consumes it first, wins!). This is despite the purpose of `.clone()`

pi0:
Thanks for headsup @MichaelDeBoey :D I saw the article in weekend and TBH, (like many other PRs) i sensed it is AI-generated but couldn't tell, also the person is AI!! 

Anyway i would say it was an _above the average_ contribution quality and it went to the very same review process as for a human.

PR #203, opened 2026-04-27
fix(node): flatten writeHead headers on Deno

### LINKED ISSUE #202: bug: Deno 2.7.13 breaks sendNodeResponse because tuple-list headers are passed to writeHead()
### Environment

srvx: 0.11.15
deno 2.7.13

### Reproduction

https://github.com/ktKongTong/srvx/tree/deno-write-header-repro

```ts
import { IncomingMessage, ServerResponse } from "node:http";
import { PassThrough } from "node:stream";
import { sendNodeResponse } from "./src/adapters/node.ts";

const socket = new PassThrough();
const req = new IncomingMessage(socket as any);
const res = new ServerResponse(req);
(res as any).assignSocket(socket);

// trigger writeHead slow path
res.setHeader("x-existing-foo", "bar");

const webRes = new Response("ok\n", {
  status: 200,
  statusText: "OK",
  headers: [
    ["x-bar", "bar"]
  ],
});

// This directly calls srvx's current implementation.
// Buggy srvx@0.11.15 on Deno v2.7.13 throws:
// TypeError: The "name" argument must be of type string. Received an instance of Array
await sendNodeResponse(res, webRes);
```

### Describe the bug

`srvx` currently passes rawHeaders to `ServerResponse.writeHead()` on Deno:

```ts
const writeHeaders: any = globalThis.Deno ? rawHeaders : rawHeaders.flat();
```

This fails on Deno 2.7.13 with writeHead() slow path and throws:
```
TypeError: The "name" argument must be of type string. Received an instance of Array
```

### Additional context

Deno v2.7.13 also causes another `deno-node-compat` CI test to fail: `cancel reading body`. I haven't addressed it yet.

### Logs

```sh

```

ktKongTong:
resolves #202

#129 added a Deno-specific branch because the Deno node-compat tests introduced in #128 exposed issues after #40.

However, a version matrix shows that flattened headers are supported by all Deno versions where tuple-list headers work, and `rawHeaders.flat()` is the only shape that works on Deno 2.7.13.

Tested across Deno versions:

| deno | rawHeaders | flatten |
|---|---:|---:|
| <= 1.41.0 | ❌ | ❌ |
| 1.41.1–2.7.12 | ✅ | ✅ |
| 2.7.13 | ❌ | ✅ |

Deno’s docs describe `rawHeaders` as a flat list, not a list of tuples:

> The keys and values are in the same list. It is not a list of tuples.

https://docs.deno.com/api/node/http/~/IncomingMessage.prototype.rawHeaders

This removes the deno-specific branch and keep same behavior with node.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency in header handling across different runtime environments.

### DIFF
--- a/src/adapters/_node/send.ts
+++ b/src/adapters/_node/send.ts
@@ -51,8 +51,7 @@ function writeHead(
   // https://github.com/nodejs/node/blob/v24.10.0/lib/_http_outgoing.js#L417
   // But it has an inconsistency in slow-path that does not unflattens!!
   // https://github.com/h3js/srvx/pull/40
-  // Deno does not support flatten in both cases.
-  const writeHeaders: any = globalThis.Deno ? rawHeaders : rawHeaders.flat();
+  const writeHeaders = rawHeaders.flat();
   if (!nodeRes.headersSent) {
     if (nodeRes.req?.httpVersion === "2.0") {
       // @ts-expect-error


ellis:
Thanks @ktKongTong - I came here to suggest the exact same fix.

pi0 (approved):
Thnx! (it seems latest deno has more regressions checking them too)

PR #205, opened 2026-05-14
fix(aws-lambda-streaming): handle empty body

### LINKED ISSUE #204: Aws Lambda Streaming responses don't work without response bodies.
### Environment

AWS Lambda Node 24

### Reproduction

N/A

### Describe the bug

When using the AWS Lambda Streaming handler, responses that don't have bodies break. 

Inside my framework, I'm trying to return a redirect response from a route:

```
return new Response(null, {
      status: 302,
      headers: {
        location: "/",
      },
    });
```

But when deployed with the preset, the lambda breaks and eventually times out.

### Additional context

_No response_

### Logs

```sh

```

taylorfsteele:
Resolves #204 

When the request body is `null` the response stream doesn't close. This breaks responses like redirects that don't have a body:
```
return new Response(null, {
   status: 302,
   headers: {
     location: "/",
   },
});
```

This PR adds in the conditional logic to create a response body if none exists.

## Release Notes

* **Bug Fixes**
  * Enhanced handling of AWS Lambda responses that lack a response body, ensuring proper stream closure and metadata delivery
  * Improved streaming behavior for responses without body content (such as redirects), maintaining correct status codes and headers in the stream

[Review Change Stack](https://app.coderabbit.ai/change-stack/h3js/srvx/pull/205)

### DIFF
--- a/src/adapters/_aws/utils.ts
+++ b/src/adapters/_aws/utils.ts
@@ -178,14 +178,17 @@ export async function awsStreamResponse(
 
   // awslambda is a global provided by Lambda runtime
   const writer = (globalThis as any).awslambda.HttpResponseStream.from(responseStream, metadata);
-
-  if (!response.body) {
-    writer.end();
-    return;
-  }
+  const body =
+    response.body ??
+    new ReadableStream<string>({
+      start(controller) {
+        controller.enqueue("");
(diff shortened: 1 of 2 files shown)


pi0 (approved):
thnx!

PR #213, opened 2026-06-23
feat: add mtls plugin

### LINKED ISSUE #212: feat: add mTLS support
### Describe the feature

In a world where CVEs are getting discovered every other day, passwords are no longer enough and post-quantum cryptography keys should become the norm it would be nice to provide direct support for mutual TLS within srvx.

In short, mTLS (mutual TLS) is a way to allow both client and server to present their respective certificates and establish a zero trust connection.

The current support is mainly focused on the Node runtime itself, Deno only recently (v2.8) added support for it only via Node compatibility (so not `Deno.serve` but `node:https`+`node:tls`), while other runtimes like Bun do not have any sort of compatibility yet (no TLSSocket), in fact it [silently accept any mTLS intermittently](https://github.com/oven-sh/bun/issues/16254) if you tell it to untrust them.

My proposal, since the API won't change that much when the first party API from both Deno and Bun will support it, is to add now the initial implementaiton for the Node part, allowing for it and Deno in node-compat to take advantage of it, and having the peer certificate available application logic.

I already have a PoC PR for this which works fine for self signed certificates, but haven't been able to also test it against any certificates issued by a trusted CA

### Additional information

- [x] Would you be willing to help implement this feature?

sandros94:
Resolves #212

As described in the issue, this exposes the client's certificate to application logic as part of the incoming request. This is mainly supported in Node and Deno in node-compat (so `srvx/node` and not `Deno.serve`), but does not work in Bun in any form (in fact, `node:tls` would trust any incoming mTLS but neither exposes it application logic nor you can force it reliably to untrust any incoming certificates).

This is a working PoC I use for personal experimentation and self-signed certificates, but I haven't tested it with those issued by trusted CA, as I do not have access to any atm.

## Summary by CodeRabbit

* **New Features**
  * Added optional mutual-TLS support for Node-based servers.
  * Exposed TLS connection details on requests, including peer certificate info, authorization status, protocol, and cipher.
  * Added a new public TLS plugin entry for easier configuration.

* **Documentation**
  * Expanded TLS and middleware guides with setup steps, usage examples, and runtime compatibility notes.

* **Tests**
  * Added coverage for TLS and mutual-TLS behavior across supported runtimes.


pi0x:
Nice work — the feature is solid where it lands. I'm holding it as a draft while we settle the API shape, because cross-runtime support is **partial** in one specific spot that carries a security footgun. Full support matrix below, all cells measured locally (Node, Deno 2.8.3, Bun 1.3.14).

### Support matrix (empirically checked)

| Runtime | Path | Enforces `requestCert`/`rejectUnauthorized` | Exposes `request.tls` peer cert |
|---|---|---|---|
| **Node** | `srvx/node` → `node:https` | ✅ | ✅ |
| **Deno** | native `srvx/deno` → `Deno.serve` | ⛔ unsupported → srvx throws loudly | ⛔ |
| **Deno** | `srvx/node` → `node:https` | ✅ | ✅ |
| **Bun** | `srvx/node` → `node:http(s)` | ✅ | ✅ |
| **Bun** | native `srvx/bun` → `Bun.serve` | ✅ | ❌ not exposed |

**On Bun**, importing `srvx/node` is a fully working mTLS workaround — enforce *and* expose — same as on Deno. I verified this with a native `curl` client (`tlsv13 alert certificate required` for a no-cert client; `{hasTls:true, authorized:true, subjectCN:"Test Client"}` for a valid one). Note that Bun's *own* HTTPS clients (`fetch({tls})` / `node:https.request`) currently fail to connect to such a server, which can make this path look broken — it isn't; the server side works. [oven-sh/bun#16254](https://github.com/oven-sh/bun/issues/16254) appears fixed for the `node:http(s)` server path.

> ⚠️ The `bun.ts` code comment currently states the opposite ("Bun's `node:http(s)` server also hands the handler a plain socket … importing `srvx/node` is not a workaround"). That's no longer accurate on Bun 1.3.14 and should be updated.

### The risk: silent auth bypass on native `Bun.serve`

The one problem cell is **native `Bun.serve`**: it *enforces* mTLS at the handshake but does **not** expose the certificate, so `request.tls` is `undefined`. A handler that authorizes on identity —

```ts
if (!request.tls?.authorized) return new Response("forbidden", { status: 403 });
```

— behaves correctly on Node/Deno/`srvx/node`-on-Bun but **silently changes meaning on native `Bun.serve`**, where `request.tls` is always `undefined`. Written slightly differently (`if (request.tls && !request.tls.authorized)`) it **fails open** — every request is treated as un-authenticated-but-allowed. Same API, opposite security outcome, no error. Shipping `request.tls` as a first-class core field invites exactly this cross-runtime bypass.

### Suggestion: ship as a srvx plugin for now

Rather than bake a security-sensitive field into the core `ServerRequest` surface while one runtime path silently no-ops it, this could live as an **opt-in srvx plugin**. That:

- keeps the core request contract uniform (no field that means different things per platform),
- makes the constraint explicit at the point of use, and can throw/warn loudly on native `Bun.serve` instead of silently returning `undefined`,
- leaves an easy path to promote it into core once `Bun.serve` exposes the peer cert and all paths are at parity.

Happy to help draft the plugin version if we agree on the direction. (For what it's worth, the landscape is friendlier than a first glance suggests — 2 of 3 runtimes work fully via `srvx/node`; only native `Bun.serve` is expose-blind.)

### Minor cleanups already pushed
Squashed the conditional TLS-option spreads in the node/bun adapters to plain spreads (both runtimes treat `undefined` options as absent) and dropped a dead type intersection — no behavior change.

---
<sub>🤖 Generated with the help of AI (Claude Code), human-reviewed. Empirical checks run locally against Node, Deno 2.8.3, and Bun 1.3.14 (mTLS probes driven with a native curl client after Bun's own HTTPS clients proved unreliable).</sub>

sandros94:
The opt-in plugin is definitely the better approach considering the upstream limitations and the unknown ETAs (for both `Deno.serve` and `Bun.serve`).

I'll need to investigate a bit as haven't built a srvx plugin, yet

sandros94:
> [oven-sh/bun#16254](https://github.com/oven-sh/bun/issues/16254) appears fixed for the `node:http(s)` server path.

Just as an heads up, but I'm not able to reproduce this claim by @pi0x.

When I try to run http2 on bun or even just `bun vitest run test/tls-peer-cert.test.ts` it always falls back to my node 24.17, making tests pass but never under bun.

IDK if I have something wrong configured in my own bun or we need to make sure to have better reproduction and tests on this topic?

PS: I'm wiring up the plugin, @pi0 WDYT about `import { mtls } from "srvx/plugins"`? The only other place I could think of would be `import { serve, mtls } from "srvx/node"`

pi0:
@sandros94 lets do try from /mtls subpath and guard plugin to hard fail (could be warn too) on unsupported runtimes. 

I guess we need to run `bun --bun` to force bun as runtime

sandros94:
> @sandros94 lets do try from /mtls subpath and guard plugin to hard fail (could be warn too) on unsupported runtimes.

Good on `/mtls` and yes I'm going for the hard fail (considering the sensitive topic)

> I guess we need to run `bun --bun` to force bun as runtime

ah true, and yes it hardly fails on providing support (http2 → `ECONNREFUSED`, http1 → `hasTls:false`)

docs/1.guide/5.options.md
```diff
@@ -94,6 +94,38 @@ serve({
 > [!TIP]
 > You can pass inline `cert` and `key` values in PEM format starting with `-----BEGIN `.
 
+### Mutual TLS (mTLS)
```
pi0: It is in options + handler docs. i am thinking maybe we can start adding a /docs/tls page for single page docs? (also notes about built in TLS)

src/mtls.ts
```diff
@@ -0,0 +1,171 @@
...
+  cipher?: TLS.CipherNameAndProtocol;
+}
+
+// Opting into the plugin augments `request.tls` onto the request. Consumers that
+// don't import `srvx/mtls` don't get the field — keeping it truly opt-in.
+//
+// `"srvx"` is the public specifier consumers resolve `ServerRequest` through, so the
+// augmentation must target it. `"./types.ts"` mirrors it for srvx's own internal
+// type-check, where `ServerRequest` is reached through the relative source module.
+declare module "srvx" {
```
pi0, line 47: Nice!

src/mtls.ts
```diff
@@ -0,0 +1,171 @@
...
+export function mtls(options: MTLSOptions = {}): ServerPlugin {
+  return (server) => {
+    // Only the Node.js adapter exposes the peer certificate. Fail loudly on anything
+    // else instead of silently leaving `request.tls` empty in production.
+    if (server.runtime !== "node") {
+      throw new Error(
+        `[srvx] mtls() requires srvx's Node.js adapter (import { serve } from "srvx/node"). The "${server.runtime}" server cannot request or expose client certificates.`,
+      );
+    }
+    if (globalThis.process?.versions?.bun) {
```
pi0: can be simplified to `"Bun" in globalThis`

docs/1.guide/5.options.md
sandros94: yes, I didn't like having that duplication of info either. I'll draft a sinlge unified `/docs/tls` page

src/mtls.ts
sandros94: > Nice!

Learnt this from opus a while back and now I cannot stop using it when developing plugins that have dedicated entries 🙈

docs/1.guide/5.options.md
sandros94: ended up adding it as page 6 (right after options) and shifting all the others, kept content simple and with basic examples

pi0 (approved):

PR #216, opened 2026-07-02
feat: `maxRequestBodySize` server option

pi0x:
Adds an opt-in `maxRequestBodySize` server option that caps the size of an incoming request body, enforced across Node, Bun, and Deno.

## Motivation

srvx had no way to bound request-body memory. The buffered read path (`request.text()` / `request.json()`) accumulated every chunk with no limit, and the streamed/native paths (`arrayBuffer()`, `blob()`, `bytes()`, `formData()`, `request.body`) had none either — so a large or malicious upload (multipart `formData()` being the most common heavy case) could grow process memory unbounded. Bun ships a native `maxRequestBodySize` (default 128 MiB); Node and Deno have no default limit.

## What this does

New generic `ServerOptions.maxRequestBodySize` (bytes, default `undefined` = no limit). When a body exceeds it, reading aborts and rejects with a `413`-style error (`statusCode: 413`, `status: 413`, `code: "ERR_BODY_TOO_LARGE"`) that a handler or `onError` can map to `413 Payload Too Large`.

The limit covers **all** read paths — buffered (`text()`/`json()`) and streamed (`request.body`, and therefore `arrayBuffer()` / `blob()` / `bytes()` / `formData()`).

### Per-runtime enforcement

| Runtime | How |
|---|---|
| **Node** | srvx wraps `request.body`'s `ReadableStream` with a pull-based byte-limiter (preserves backpressure, cancels upstream on overflow); the buffered `readBody()` fast path keeps its own cheap check. |
| **Bun** | forwarded to Bun's native `maxRequestBodySize` — Bun responds `413` before the handler runs. An explicit `bun.maxRequestBodySize` still wins. |
| **Deno** | `Deno.serve` has no native option, so srvx rebuilds the request with a size-limited body stream (method/url/headers/**signal** preserved). |

The shared limiter primitives live in a runtime-agnostic `src/_body-limit.ts`. Also threaded through `toNodeHandler(handler, { maxRequestBodySize })` for handlers embedded in an existing Node server.

## Hot path / backward compat

Default `undefined` → **no limit**, identical to today. When unset, no body-stream wrapping and no per-request allocation occurs on any runtime (a separate commit also reverted an incidental per-request object allocation so the Node default path matches `main` exactly, ~25 ns/ctor).

## Tests

- `test/node-max-body-size.test.ts` — 9 cases: per-method (`text`/`json`/`arrayBuffer`/`bytes`/`blob`/`formData`/streaming) over-limit → 413, within-limit → 200, and no-limit-by-default backward-compat.
- `test/bun-max-body-size.test.ts` + `test/deno-max-body-size.test.ts` — spawn the target runtime against `test/fixtures/max-body-server.ts` and assert 413 over / 200 within.

All suites green (node/bun/deno adapter suites + body-size tests: 191 passed, 8 skipped). `pnpm lint` and `pnpm typecheck` clean. Docs updated in `docs/1.guide/5.options.md`.

## Notes for review

- **Bun semantics differ slightly**: Bun rejects at the protocol level (413, empty body, handler never runs), whereas Node/Deno surface `ERR_BODY_TOO_LARGE` as a catchable rejection. Inherent to using Bun's native option.
- **Deno rebuilds the `Request`** per bodied request when a limit is set (opt-in only) — a small allocation on that path; happy to gate it further if preferred.
- Option name matches Bun's `maxRequestBodySize` for consistency; unreleased, so no alias kept.

## Summary by CodeRabbit

* **New Features**
  * Added an optional request body size limit for Node.js servers.
  * Large buffered request bodies can now be rejected with a clear 413-style error when they exceed the configured limit.

* **Bug Fixes**
  * Improved handling of request body buffering so oversized payloads stop reading sooner.
  * Existing behavior remains unchanged when no limit is set.

### DIFF
--- a/docs/1.guide/5.options.md
+++ b/docs/1.guide/5.options.md
@@ -123,6 +123,41 @@ serve({
 });
 ```
 
+### `maxRequestBodySize`
+
+Maximum allowed size **in bytes** for the request body. Defaults to `undefined` (no limit).
+
+As the body is read, its accumulated length is tracked and, once it exceeds the limit, reading is aborted and rejects with a `413`-style error. The error carries `statusCode: 413`, `status: 413` and `code: "ERR_BODY_TOO_LARGE"`, so a handler (or [`onError`](#onerror)) can map it to an HTTP `413 Payload Too Large` response.
+
+The limit covers both buffered reads (`request.text()` / `request.json()`) and the streamed body (`request.body`, and therefore `request.arrayBuffer()` / `.blob()` / `.bytes()` / `.formData()`).
+
+**Example:**
+
+```js
+import { serve } from "srvx";
+
+serve({
+  maxRequestBodySize: 1024 * 1024, // 1 MiB
+  fetch: async (request) => {
+    try {
+      return Response.json(await request.json());
+    } catch (error) {
+      if (error.code === "ERR_BODY_TOO_LARGE") {
+        return new Response("Payload Too Large", { status: 413 });
+      }
+      throw error;
+    }
+  },
+});
+```
+
+> [!NOTE]
+> Runtime support:
+>
+> - **Node**: enforced by srvx (the request body stream is size-limited).
+> - **Bun**: forwarded to Bun's native [`maxRequestBodySize`](https://bun.sh/docs/api/http), enforced by Bun (responds with `413` before the handler runs).
+> - **Deno**: enforced by srvx (`Deno.serve` has no native option).
+
 ## Runtime Specific Options
 
 ### Node.js
--- a/src/_body-limit.ts
+++ b/src/_body-limit.ts
@@ -4,24 +4,27 @@
 
 /**
  * Creates a `413 Payload Too Large` style error for when a request body exceeds
- * the configured `maxBodySize`. The `statusCode` / `status` properties let a
+ * the configured `maxRequestBodySize`. The `statusCode` / `status` properties let a
  * handler map it to an HTTP 413 response.
  */
-export function createBodyTooLargeError(maxBodySize: number): Error {
+export function createBodyTooLargeError(maxRequestBodySize: number): Error {
   return Object.assign(
-    new Error(`Request body exceeds the maximum allowed size of ${maxBodySize} bytes.`),
+    new Error(`Request body exceeds the maximum allowed size of ${maxRequestBodySize} bytes.`),
     { code: "ERR_BODY_TOO_LARGE", statusCode: 413, status: 413 },
   );
 }
 
 /**
  * Wraps a body `ReadableStream` so the total number of bytes read cannot exceed
- * `maxBodySize`. Once the limit is passed the wrapped stream errors with a
+ * `maxRequestBodySize`. Once the limit is passed the wrapped stream errors with a
  * `413`-style error and the upstream stream is cancelled. This is pull-based, so
  * it preserves backpressure and stops reading as soon as the limit is hit rather
  * than buffering the whole body first.
  */
-export function limitBodyStream(stream: ReadableStream, maxBodySize: number): ReadableStream {
+export function limitBodyStream(
+  stream: ReadableStream,
+  maxRequestBodySize: number,
+): ReadableStream {
   const reader = stream.getReader();
   let size = 0;
   return new ReadableStream({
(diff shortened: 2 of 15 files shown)


pi0x:
### Follow-up: hot-path allocation fix + Bun/Deno default comparison

**Perf fix (pushed):** The initial commit changed `runtime.node` from reusing the passed context object to allocating a fresh `{ req, res }` literal per request (to keep `maxBodySize` off the public shape). Benchmarking showed this added ~4–5 ns to **every** request on the hot path, even with no limit configured. The follow-up commit reverts to `node: ctx` — `maxBodySize` rides along but is intentionally excluded from the public `runtime.node` type, and all consumers only read `req`/`res`. Measured constructor cost is now ~25 ns/ctor, identical to `main` (within noise). The per-chunk `readBody` guard was already correctly skipped when `maxBodySize` is unset, so the default path is now truly zero-overhead.

**Do other runtimes impose a default?**

| Runtime | Default request-body limit |
|---|---|
| **Bun** (`Bun.serve`) | **128 MiB** (`maxRequestBodySize`, 134,217,728 bytes) |
| **Deno** (`Deno.serve`) | **None** (open feature request [denoland/deno#3515](https://github.com/denoland/deno/issues/3515)) |
| **Node** (raw `http`) | **None** (only `maxHeaderSize` ~16 KB for headers) |

So there's precedent (Bun) for a safe default, but Deno and raw Node have none. Left as `undefined` (opt-in) here because this limit currently only covers the **buffered** `.text()`/`.json()` fallback — streaming (`request.body`) is unaffected — so a default-on limit would apply asymmetrically and could surprise streaming apps. If a default is desired for parity with Bun, 128 MiB would match, but I'd suggest deciding that separately from this guard mechanism.

pi0x:
### Update: `maxBodySize` now covers all body-read paths + `toNodeHandler`

The initial commit only enforced the limit in the buffered `readBody()` fallback (`text()`/`json()`). Reads that route through the native `Request` — `arrayBuffer()`, `blob()`, `bytes()`, `formData()` — and direct streaming of `request.body` bypassed it, which meant the most common heavy path (multipart file uploads via `formData()`) was uncapped.

`c41c1bc` closes that gap by enforcing at the single choke point every consumer funnels through: `request.body`'s `ReadableStream` is wrapped with a **pull-based** limiter that counts bytes and errors with the same `413`-style error (`ERR_BODY_TOO_LARGE`, `statusCode: 413`) once exceeded, cancelling the upstream Node stream. Because it's pull-based it preserves backpressure and stops reading as soon as the limit is hit rather than buffering first. The `readBody()` fast path keeps its own cheap check.

Verified end-to-end that `arrayBuffer()`, `bytes()`, `blob()`, `formData()`, streaming, and `text()`/`json()` all return **HTTP 413** with the limit set, and within-limit requests still succeed. Tests expanded to 9 (per-method + multipart + within-limit + backward-compat); full node suites pass (145 passed / 6 skipped).

**Hot path unchanged:** when `maxBodySize` is `undefined` (default) the body stream is not wrapped — no overhead. Combined with `8f2bc6a` (reuse the request context object), the default path has zero added allocation vs `main`.

Also threaded through `toNodeHandler(handler, { maxBodySize })` so handlers embedded in an existing Node server can opt in. Still Node-only; other runtimes ignore the option (docs updated to say "Currently only enforced for the Node.js runtime").

pi0x:
### Update: `maxBodySize` is now a cross-runtime option (Bun + Deno)

Promoted `maxBodySize` from Node-only to a generic server option, enforced on every runtime srvx can enforce it on:

- **Extracted** the body-limit primitives into a runtime-agnostic module `src/_body-limit.ts` — `createBodyTooLargeError`, `limitBodyStream` (used by the Node body getter), and `limitRequestBody` (rebuilds a native `Request` with a size-limited body, preserving method/url/headers/**signal**).
- **Bun**: mapped to Bun's native `maxRequestBodySize` — Bun returns **413 before the handler runs**. An explicit `bun.maxRequestBodySize` still wins. (Bun's own default is 128 MiB; srvx leaves it unset unless you set `maxBodySize`.)
- **Deno**: `Deno.serve` has no native body-size option, so srvx enforces it in the fetch wrapper by rebuilding the request with a limited body stream. No-op when unset or bodyless, so the default hot path is unchanged.
- **Node**: unchanged (srvx stream limiter + fast-path `readBody` check).

Docs updated to a per-runtime support table.

**Verified end-to-end on all three runtimes** (413 over the limit, 200 within), with automated tests: Node via vitest (9 cases), and Bun/Deno via spawned fixtures (`test/bun-max-body-size.test.ts`, `test/deno-max-body-size.test.ts`, sharing `test/fixtures/max-body-server.ts`). Full suites green: node/bun/deno adapter suites + all body-size tests (191 passed / 8 skipped).

pi0 (approved):

PR #229, opened 2026-07-14
fix: resolve `x-forwarded-*` hop-aware, right-to-left

pi0x:
The `X-Forwarded-*` headers were resolved **leftmost-first**, so under the recommended `trustProxy: "loopback"` an attacker sending `X-Forwarded-For: 9.9.9.9` got `request.ip === "9.9.9.9"`. Because real proxies **append** to the header, the leftmost value is fully client-controlled — this defeats IP rate-limits, allowlists, geo checks and audit logs (F7). The same class of issue let forwarded proto and host be poisoned (cache keys, password-reset links).

## New hop-aware model (D1)

Resolve the forwarded chain **hop-aware, right-to-left** from the immediate peer, skipping addresses in the trusted set — the first untrusted address is the client. This matches Express (`trust proxy`), nginx (`$proxy_add_x_forwarded_for`) and AWS ALB.

- The client is the **rightmost address _not_ in the trusted set**.
- If every hop is trusted, the leftmost entry is the client.
- If the immediate peer is untrusted, the forwarded headers are ignored entirely.
- The same model gates forwarded **proto/host**: only the value contributed by the trusted proxy is honored, so a spoofed prefix cannot poison `request.url`.

## Changes

- Replace `firstForwardedValue` with `forwardedList` / `resolveClientIP` / `trustedHops` / `forwardedHopValue` in `src/_trust-proxy.ts`.
- Apply at all three call sites: universal plugin, Node adapter (`_node/request.ts` + `_node/url.ts`), and AWS adapter (`_aws/utils.ts`, starting from the gateway-verified `sourceIp` as the nearest hop).
- Rewrite the tests that enshrined the old leftmost behavior and add hop-aware cases (spoofed-prefix, all-trusted, untrusted-peer, proto/host poisoning).
- Rewrite the `trustProxy` docs to state the hop model explicitly (Docs-T1).

## Verification

- `pnpm vitest run --exclude '**/{deno,bun}*.test.ts' test/` — 860 passed, 32 skipped
- `pnpm lint` and `pnpm typecheck` clean

### DIFF
--- a/docs/1.guide/5.options.md
+++ b/docs/1.guide/5.options.md
@@ -160,12 +160,23 @@ serve({
 
 Whether to trust `X-Forwarded-*` headers (`X-Forwarded-Proto`, `X-Forwarded-Host`, `X-Forwarded-For`, and the HTTP/2 `:scheme`) when deriving `request.url` and `request.ip`. Defaults to `false`.
 
-Any client can send `X-Forwarded-Proto: https`, `X-Forwarded-Host` or `X-Forwarded-For`, so trusting them lets a request masquerade as `https:`, forge its host, or spoof its client IP. Only enable this when a proxy you control sits in front and overwrites the headers.
+Any client can send `X-Forwarded-Proto: https`, `X-Forwarded-Host` or `X-Forwarded-For`, so trusting them lets a request masquerade as `https:`, forge its host, or spoof its client IP. Only enable this when a proxy you control sits in front.
 
 - `false` (default): ignore the headers; use the real connection protocol, the on-the-wire `Host` header and the socket peer address.
-- `true`: always trust the headers.
-- `"loopback"`: trust them only when the proxy connects from a loopback address (`127.0.0.0/8` or `::1`).
-- `string[]`: trust them only when the proxy's address is in the list.
+- `true`: trust every hop (the leftmost `X-Forwarded-For` entry is the client).
+- `"loopback"`: trust hops on a loopback address (`127.0.0.0/8` or `::1`), i.e. a proxy on the same host.
+- `string[]`: trust hops whose address is in the allowlist.
+
+#### Hop-aware resolution
+
+Real proxies **append** to `X-Forwarded-*`, so the chain grows left-to-right (outermost client first, nearest proxy last). srvx resolves it **hop-aware, right-to-left**, matching Express (`trust proxy`), nginx (`$proxy_add_x_forwarded_for`) and AWS ALB:
+
+- Starting from the immediate peer, each address in the trusted set is treated as a proxy you control.
+- **The client is the rightmost address _not_ in the trusted set.**
+- If every address in the chain is trusted, the leftmost entry is the client.
+- If the immediate peer is not trusted, the headers are ignored entirely.
+
+This is why a spoofed prefix cannot poison `request.ip`: with `trustProxy: "loopback"` and a request carrying `X-Forwarded-For: 9.9.9.9`, the trusted proxy appends the real client address, so the forged `9.9.9.9` is skipped rather than returned. The same model gates the forwarded protocol and host: only the value contributed by the trusted proxy is honored, so an attacker cannot poison `request.url` (e.g. cache keys or password-reset links).
 
 **Example:**
 
--- a/src/_trust-proxy.ts
+++ b/src/_trust-proxy.ts
@@ -4,16 +4,19 @@ import type { ServerPlugin, ServerRequest } from "./types.ts";
  * Controls whether `X-Forwarded-*` headers (proto, host, for, and the HTTP/2
  * `:scheme` pseudo-header) are trusted when deriving request metadata.
  *
- * These headers are set by the client on the wire, so they can only be trusted
- * when a proxy you control sits in front and overwrites them. See
- * {@link ServerOptions.trustProxy}.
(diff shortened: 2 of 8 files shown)


pi0x (commented):
## Adversarial review — F7 / D1 hop-aware `X-Forwarded-*`

**Verdict: Approve with comments.** The spec (D1 / F7) is fully met and I could not construct a spoof that works under the recommended `trustProxy: "loopback"` or an allowlist of exact proxy IPs. The right-to-left walk is correct and consistent across all three call sites. The findings below are edge cases and pre-existing limitations, not spec violations — the top one is a low-severity wart introduced here that slightly undercuts the PR's own "spoofed prefix cannot poison `request.url`" claim.

### Spec-compliance checklist
- [x] XFF resolved hop-aware, right-to-left; first untrusted = client — `resolveClientIP` (`src/_trust-proxy.ts:116-131`). Verified vs Express/nginx semantics.
- [x] `firstForwardedValue` gone — no references remain in `src/`, `test/`, `docs/`.
- [x] Same hop model for forwarded **proto** and **host** — `forwardedHopValue` + `trustedHops` (`_trust-proxy.ts:146-187`), robust against prepended entries (selects from the right via `length - hops`).
- [x] Applied to universal plugin (`_trust-proxy.ts:209-258`) **and** Node (`_node/request.ts:83-108`, `_node/url.ts`).
- [x] AWS half: `awsEventClientIP` removed; `awsRequest` now starts from the gateway-verified `sourceIp` and runs `resolveClientIP` / `trustedHops` (`_aws/utils.ts:40-55`). No more blind leftmost override.
- [x] Old enshrined tests rewritten (not just kept green): `node-trust-proxy.test.ts:207`, `trust-proxy-plugin.test.ts:118` now assert `trustProxy:true`→leftmost with explicit rationale, plus new spoofed-prefix / all-trusted / untrusted-peer / proto-host cases.
- [x] Docs state the hop model explicitly (client = rightmost address not in the trusted set) — `docs/1.guide/5.options.md`.
- Tests: `node-trust-proxy` + `trust-proxy-plugin` + `aws` = 101 passed. Full non-deno/bun suite: 859 passed, 32 skipped, 1 fail in `test/loader.test.ts` which **also fails on the base commit** (pre-existing, unrelated).

### Findings (by severity)

**1. [Low] Absent `X-Forwarded-For` + trusted peer ⇒ `trustedHops` returns `Infinity` ⇒ proto/host fall back to the *leftmost* (most attacker-controllable) entry.** `src/_trust-proxy.ts:154-163`.
When the XFF list is empty but the peer is trusted, the loop finds no untrusted boundary and returns `Number.POSITIVE_INFINITY`; `forwardedHopValue(list, Infinity)` then clamps to `list[0]` (leftmost). This contradicts the PR's own anti-leftmost thesis for proto/host.
Scenario: a trusted proxy that appends `X-Forwarded-Proto`/`-Host` but does **not** set `X-Forwarded-For` (unusual but real). Confirmed locally:
`forwardedHopValue("https, http", trustedHops("loopback","127.0.0.1", undefined))` → `"https"` — the attacker-prepended scheme wins over the proxy's real `http`.
Not exploitable under a normal loopback nginx (`$proxy_add_x_forwarded_for` always sets XFF), hence Low. Safer fix: treat an absent/empty XFF as a single peer hop (rightmost) rather than unbounded — i.e. don't return `Infinity` when the list is empty. Worth at least a code comment acknowledging the trade-off, since it's the one place the model degrades toward the behavior this PR exists to remove.

**2. [Low, pre-existing — now broader impact] No CIDR / preset-range support in the trusted set.** `src/_trust-proxy.ts:45-53` (`isTrustedProxy` uses exact `Array.includes`).
`trustProxy: ["10.0.0.0/8"]` — a very natural thing to write — matches nothing: `trustedHops(["10.0.0.0/8"],"10.0.0.1","1.2.3.4")` → `0`, so all forwarded headers are silently ignored **and** `request.ip` becomes the proxy address `10.0.0.1`. This PR now runs the matcher against *every* XFF hop, so an intermediate proxy inside a subnet stops the walk early and is returned as the client. Express `trust proxy` accepts subnets and preset names (`loopback`/`linklocal`/`uniquelocal`); the docs here invite `string[]` without stating "exact IPs only". Recommend documenting the exact-match constraint now, and considering CIDR support (additive later).

**3. [Low, pre-existing] XFF entries carrying a port or bracketed IPv6 don't match a bare allowlist entry.** `[2001:db8::1]:443` vs `["2001:db8::1"]` → treated as untrusted and returned as the client (confirmed). No IPv6 canonicalization (`::1` vs long form). Minor because XFF is normally bare IPs, but some proxies append `ip:port`.

**4. [Informational] Docs slightly overstate the protection.** The "a spoofed prefix cannot poison `request.url`" guarantee holds only when the trusted proxy *appends/overwrites* proto and host. A trusted peer that merely passes client-supplied `X-Forwarded-Host`/`-Proto` through still allows poisoning — inherent to the trust model (same caveat as Express), but the new prose reads as an unconditional guarantee.

### Nice work
The `length - hops` selection for proto/host correctly tracks the same index as the client's XFF entry in a clean appending chain, and degrades safely (toward the peer/rightmost) when an attacker prepends extra proto/host entries. The per-request memoization in the Node adapter (`#remoteAddr`/`#resolveHops`) is clean and keeps `ip` and `_url` consistent.

pi0 (approved):

PR #230, opened 2026-07-14
feat: v1 api freeze

pi0x:
API stabilization batch for v1. All breaking-shaped changes land together.

- **F8** (breaking-shaped): wire `errorPlugin` on the Bun and Deno adapters so the `error` option runs on every adapter; docs section renamed `onError` -> `error` to match the real option.
- **F9** (breaking-shaped): remove the unimplemented `gracefulShutdown.forceTimeout` type; fix the falsy check so `gracefulTimeout: 0` is honored; document `gracefulShutdown` (unit is **seconds**, default `5`, `SERVER_SHUTDOWN_TIMEOUT` env var).
- **F10** (breaking-shaped): standardize `Server.serve()` on `Promise<Server<Handler>>`; adapters that returned `void` now return `Promise.resolve(this)`.
- **F11**: drop `@experimental` from `toFetchHandler` / `fetchNodeHandler` (stable/frozen at v1); `srvx/tracing` stays experimental.
- **F17** (breaking-shaped): `NodeResponse` now throws (like native `Response`) for a non-null body with a null-body status (`101/204/205/304`) instead of emitting an illegal body + content-length.
- **F26 / F18** (breaking-shaped): Bun/Deno `serve()` never throws; listen errors (e.g. EADDRINUSE) surface via `ready()`, matching Node. If `ready()` is never awaited, the error is re-surfaced as an unhandled rejection (Node's default handler reports it) rather than leaving a live process with a dead server.
- **F51**: lazily initialize `request.context` (`??= {}`) once in shared code so `request.context.x = ...` works on every adapter.
- **F52** (breaking-shaped): remove the unused, `any`-typed `netlify` / `stormkit` / `vercel` runtime context keys (removal after v1 would itself be breaking).
- **D6**: keep `ServerPlugin` return type sync-only (`void`), not widened to `Promise<void> | void`.

Adds node-adapter tests for each behavioral change (`error` runs, `gracefulTimeout: 0` honored, `serve()` returns a `Promise<Server>`, `NodeResponse` throws on null-body-status + body, `ready()` rejects on EADDRINUSE, `request.context` works). Bun/Deno-only behaviors are covered by the existing bun/deno suites.

Verified: `pnpm vitest run --exclude '**/{deno,bun}*.test.ts' test/` (860 passed / 32 skipped), `pnpm lint`, `pnpm typecheck` all green.

### DIFF
--- a/docs/1.guide/5.options.md
+++ b/docs/1.guide/5.options.md
@@ -98,13 +98,13 @@ Client certificates (mutual TLS) are available through the [`mtls()` plugin](/gu
 
 :read-more{to="/guide/tls" title="TLS & mutual TLS"}
 
-### `onError`
+### `error`
 
-Runtime agnostic error handler.
+Runtime agnostic error handler. It runs whenever the `fetch` handler (or any middleware) throws or rejects, and returns the `Response` to send instead.
 
 > [!NOTE]
 >
-> This handler will take over the built-in error handlers of Deno and Bun.
+> This handler will also take over the built-in error handlers of Deno and Bun.
 
 **Example:**
 
@@ -113,7 +113,7 @@ import { serve } from "srvx";
 
 serve({
   fetch: () => new Response("👋 Hello there!"),
-  onError(error) {
+  error(error) {
     return new Response(`<pre>${error}\n${error.stack}</pre>`, {
       headers: { "Content-Type": "text/html" },
     });
@@ -125,7 +125,7 @@ serve({
 
 Maximum allowed size **in bytes** for the request body. Defaults to `undefined` (no limit).
 
-As the body is read, its accumulated length is tracked and, once it exceeds the limit, reading is aborted and rejects with a `413`-style error. The error carries `statusCode: 413`, `status: 413` and `code: "ERR_BODY_TOO_LARGE"`, so a handler (or [`onError`](#onerror)) can map it to an HTTP `413 Payload Too Large` response.
+As the body is read, its accumulated length is tracked and, once it exceeds the limit, reading is aborted and rejects with a `413`-style error. The error carries `statusCode: 413`, `status: 413` and `code: "ERR_BODY_TOO_LARGE"`, so a handler (or [`error`](#error)) can map it to an HTTP `413 Payload Too Large` response.
 
 The limit covers both buffered reads (`request.text()` / `request.json()`) and the streamed body (`request.body`, and therefore `request.arrayBuffer()` / `.blob()` / `.bytes()` / `.formData()`).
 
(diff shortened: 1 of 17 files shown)


pi0x (commented):
## Independent review — v1 API freeze (PR #230)

**Verdict: Approve (with one minor follow-up).** Every locked hard spec is correctly implemented and I verified each one end-to-end (unit tests + manual EADDRINUSE/null-body repros run on **Node, Bun and Deno**). Full suite is green: `923 passed | 35 skipped`, no type errors. The single failing test in a fresh checkout (`test/loader.test.ts`) is purely environmental — it needs a built `dist/` for the `import("srvx/node")` self-import and passes after `pnpm build`; `loader.ts` is untouched by this PR.

### Per-finding compliance checklist

| Finding | Status | Evidence |
|---|---|---|
| **F8/D2** — option stays `error`, actually wired on every adapter | PASS | `errorPlugin(this)` runs in node/bun/deno/cloudflare/aws-lambda/generic/bunny/service-worker, always *before* `wrapFetch`. Verified a thrown handler is caught and mapped (500 "handled: boom"). Docs `5.options.md` rewritten `### onError` → `### error`; body-limit cross-ref updated. (The remaining `onError` at 5.options.md:263 is Deno's *native* `deno.onError` option — correct.) |
| **F9/D3** — seconds, `0` honoured, `forceTimeout` deleted, documented | PASS* | `_plugins.ts:27` uses `== null` not `||`; `{gracefulTimeout: 0}` force-closes immediately (verified). `forceTimeout` removed from `types.ts` and absent everywhere. Docs state unit=seconds, default 5, `SERVER_SHUTDOWN_TIMEOUT`. *One minor gap below.* |
| **F10** — `serve(): Promise<Server<Handler>>` | PASS | `types.ts:280` narrowed; every void adapter returns `Promise.resolve(this)` / `Promise.resolve().then(() => this)`. |
| **F11/D5** — drop `@experimental` on `toFetchHandler`/`fetchNodeHandler`; tracing stays experimental | PASS | Tags removed from `_node/adapter.ts` + `_node/web/fetch.ts`; `@experimental` remains only on `src/tracing.ts`. |
| **D6** — `ServerPlugin` stays sync `void` | PASS | `types.ts:47` `(server: Server) => void`, not widened. |
| **F17** — `NodeResponse` throws for body + null-body status | PASS | Throws for 101/204/205/304 whenever `body != null` (covers strings *and* streams). Matches native `Response` (confirmed empty-string and stream bodies both throw natively). |
| **F18** — unawaited `ready()` still surfaces listen error | PASS | `reportUnhandledListenError` re-surfaces via a deferred unhandled rejection unless `ready()` claimed it. Repro: second server on a busy port, `ready()` never awaited → EADDRINUSE surfaced loudly (Node exits 1; Bun prints), *not* swallowed. |
| **F26** — `serve()` never throws, `ready()` rejects (Bun+Deno) | PASS | `Bun.serve`/`Deno.serve` wrapped; verified on **both** runtimes: `serve()` did not throw synchronously, `ready()` rejected (EADDRINUSE / AddrInUse). No spurious unhandled rejection when `ready()` is awaited. Knock-on checked: CLI (`cli/serve.ts:81`) and all docs/examples read `server.url` only after `await ready()` — no sync `server.url` regressions. |
| **F51** — `request.context` lazy-init in shared code | PASS | `??= {}` lives once in `wrapFetch` (both branches), which every adapter routes through. `req.context.user = "alice"` works (verified); native Bun/Deno Requests are extensible so assignment succeeds there too. |
| **F52** — remove `netlify`/`stormkit`/`vercel` any-typed keys | PASS | Removed from `ServerRuntimeContext`; no dangling references in `src/`. |

### Findings

**Minor — `SERVER_SHUTDOWN_TIMEOUT=0` env var is silently ignored (docs/behavior mismatch).** `src/_plugins.ts:28` — `Number.parseInt(process.env.SERVER_SHUTDOWN_TIMEOUT || "") || 5`. The `|| 5` is the exact falsy bug F9 fixed for the *option*, still present on the *env* path: `SERVER_SHUTDOWN_TIMEOUT=0` resolves to `5`, not `0`. The docs this PR adds say the timeout "can also be set via the `SERVER_SHUTDOWN_TIMEOUT` environment variable" and "Set it to `0` to force close immediately", so a user setting `0` via env gets a 5s wait instead. Suggest `const env = process.env.SERVER_SHUTDOWN_TIMEOUT; const fallback = env == null || env === "" ? 5 : Number.parseInt(env);` (or apply the same `?? / == null` guard). Not blocking — the primary option path is correct.

**Info — PR branch is stale.** Base is `f2edab1`; `main` is now `d8523dc` (#225/#227/#228/#234 landed since). Everything applies cleanly and tests pass, but a rebase before merge is advisable so CI runs against current main.

**Info (out of scope, flagging for the freeze) — `docs/1.guide/3.server.md:40-48` still documents `server.addr` / `server.port`, which don't exist** on `Server` or any adapter. That's F54 (Docs scope, not this PR), but it's semver-adjacent doc debt worth clearing before the freeze tag.

Nice touch: the added `test/v1-api-freeze.test.ts` covers error-wiring, lazy context, promise return, EADDRINUSE-via-ready, null-body throw, and the `gracefulTimeout: 0` regression directly — good behavioral (not just type) coverage that pins these decisions.

PR #233, opened 2026-07-14
fix(static): harden srvx/static

pi0x:
Hardens `srvx/static` (on the CLI default path, previously untested):

- **Decoding**: percent-decode the pathname so filenames with spaces/unicode are reachable; malformed sequences fall through (no crash). Traversal checks run on the decoded path — raw `../` and encoded `%2e%2e` stay blocked.
- **Dotfiles**: deny any path segment starting with `.` (covers `.env`, `.env.local`, `.npmrc.bak`, `.git/...`, and dot-segment traversal).
- **Symlinks**: resolve `fs.realpath` and reject files that escape `dir`.
- **Compression negotiation**: only compress compressible MIME types (never re-encode png/jpg/woff/etc.); parse `Accept-Encoding` by exact tokens + q-values (`br;q=0` disabled, `abbr` no longer matches `br`, identity fallback); set `Vary: Accept-Encoding` on both compressed and identity variants.
- **Caching / conditional requests**: add `ETag` (size+mtime) and `Last-Modified`; handle `If-None-Match` / `If-Modified-Since` → `304`; add a configurable conservative `Cache-Control` default.
- **HEAD**: sends the same headers as GET with no body/compression work.
- **Tests**: new `test/static.test.ts` covering traversal, decoding, MIME, encoding negotiation, HEAD, index/`.html` resolution, 404 fallthrough, dotfiles, symlink escape, conditional requests, `Vary`, and already-compressed types.

`Range`/206 support is intentionally deferred to a follow-up.

## Summary by CodeRabbit

* **New Features**
  * Enhanced static file serving with improved MIME type detection and optional `Cache-Control` settings.
  * Added gzip and Brotli compression negotiation based on client preferences.
  * Added caching validators and conditional responses using `ETag`, `Last-Modified`, and `304 Not Modified`.
  * Added support for `index.html`, extensionless HTML paths, and `.well-known` resources.

* **Bug Fixes**
  * Strengthened protection against path traversal, dotfiles, and unsafe symlinks.
  * Improved `HEAD` request handling and malformed URL processing.

* **Documentation**
  * Expanded static-file serving documentation, including caching, compression, routing, and security behavior.

### DIFF
--- a/docs/1.guide/10.cli.md
+++ b/docs/1.guide/10.cli.md
@@ -102,7 +102,9 @@ npx srvx --static ./dist
 
 If `--static` is omitted, srvx serves files from a `public/` directory when one exists. When both a server entry and a static directory are present, static files take priority and unmatched requests fall through to your handler.
 
-Static serving includes automatic `index.html` resolution, `.html` extension fallback (e.g. `/about` → `about.html`), common MIME types, gzip/Brotli compression, and path-traversal protection.
+Static serving includes automatic `index.html` resolution, `.html` extension fallback (e.g. `/about` → `about.html`), common MIME types, gzip/Brotli compression, and `ETag` / `Last-Modified` validators with `304` conditional-request handling.
+
+Requests are also hardened against path traversal and symlinks that escape the served directory. Dotfiles are never served (`.env`, `.env.local`, `.git/...`), with one exception: [`/.well-known/`](https://datatracker.ietf.org/doc/html/rfc8615) stays reachable so ACME/Let's Encrypt challenges, `security.txt`, and `assetlinks.json` work. Paths under `/.well-known/` are served verbatim, without the `.html` fallback.
 
 For programmatic usage, import the [`serveStatic`](https://github.com/h3js/srvx/blob/main/src/static.ts) middleware:
 
--- a/src/static.ts
+++ b/src/static.ts
@@ -2,7 +2,7 @@ import type { ServerMiddleware } from "./types.ts";
 import type { Transform } from "node:stream";
 
 import { extname, join, resolve, sep } from "node:path";
-import { readFile, stat } from "node:fs/promises";
+import { readFile, stat, realpath } from "node:fs/promises";
 import { createReadStream, ReadStream } from "node:fs";
 import { FastResponse } from "srvx";
 import { createGzip, createBrotliCompress } from "node:zlib";
@@ -18,6 +18,15 @@ export interface ServeStaticOptions {
    * The HTTP methods to allow for serving static files.
    */
   methods?: string[];
+
+  /**
+   * Value for the `Cache-Control` response header.
+   *
+   * Defaults to a conservative `"public, max-age=0, must-revalidate"` which
+   * lets clients cache but forces revalidation (via `ETag`/`Last-Modified`)
+   * on every request. Set to `false` to omit the header entirely.
+   */
+  cacheControl?: string | false;
 
   /**
    * A function to modify the HTML content before serving it.
@@ -39,6 +48,7 @@ const COMMON_MIME_TYPES: Record<string, string> = {
   ".json": "application/json",
   ".txt": "text/plain",
   ".xml": "application/xml",
+  ".wasm": "application/wasm",
   ".gif": "image/gif",
   ".ico": "image/vnd.microsoft.icon",
   ".jpeg": "image/jpeg",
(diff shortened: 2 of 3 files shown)


pi0x (commented):
## Adversarial review — `fix(static): harden srvx/static` (F31 + F32)

**Verdict: APPROVE with comments.** I could not construct a traversal or dotfile bypass, and every locked-scope spec point is implemented and tested. All 38 new tests pass; the full suite (953 passed / 35 skipped) is green; `oxlint` clean. One substantive usability finding (`.well-known`) and a few nits below — none is a security bypass or a spec violation, so this is a comment, not a block.

I checked out the PR, read `src/static.ts` + `test/static.test.ts` in context, and ran ~25 local attack probes (not committed) driving the middleware with **raw origin-form targets that bypass `Request`'s WHATWG normalization** (i.e. the real Node-adapter wire path), which the shipped tests do not exercise since they build `new Request(...)`.

### Spec checklist (F31 + F32, locked scope)
- [x] **Percent-decoding** — `decodeURIComponent` added (static.ts:177), dot-segment + `startsWith(dir)` checks kept. `hello%20world.txt` / `caf%C3%A9.txt` reachable.
- [x] **Traversal safe** — verified raw `../`, `%2e%2e`, `..%2f`, double-encoded `%252e%252e`, mixed `foo%2f..%2f..`, encoded backslash `..%5c`, and `%00`-prefixed all → 404, no secret leak. Two independent layers: the URL parser collapses/clamps *literal* `..` at root, and the middleware dot-check (static.ts:186) catches *decoded* encoded forms. Double-encoding is safe because a single decode leaves literal `%2e`/`%2f` text (no real separators).
- [x] **Null bytes / malformed percent** — `decodeURIComponent("%ZZ")` caught → `next()` (no 500). `stat()` on a null-byte path **rejects** (async), caught by `.catch(()=>null)` → 404. No sync-throw 500. Confirmed.
- [x] **Dotfile denylist** — deliberate `segment.startsWith(".")` blocks `.env`, `.env.local`, `.npmrc.bak`, `.git/...`, and dot-directories. (See finding 1 re: over-breadth.)
- [x] **Symlinks** — `realpath` on both base and file; escape rejected (static.ts:212-222). Verified with a symlinked **file** *and* a symlinked **directory** pointing outside `dir` → both 404.
- [x] **No re-compression** — `isCompressible()` gates it; `.png/.jpg/.mp4/.zip/.woff2` never encoded, and `Vary` correctly omitted for them.
- [x] **`accept-encoding` parsing** — real q-value parse; `br;q=0` → no brotli (falls to gzip/identity); `abbr` no longer matches `br`; `Vary: Accept-Encoding` on **both** compressed and identity variants of compressible types.
- [x] **Conditional / caching** — `ETag` (weak, size+mtime) + `Last-Modified`; `If-None-Match` (incl. `*`, multi-tag, `W/` normalization) precedes `If-Modified-Since`; 304 strips body/Content-Length/Content-Encoding; sane `Cache-Control` default, `false` to disable.
- [x] **HEAD** — no read/compress; same headers, null body.
- [x] **Range/206 OUT** — correctly absent. Nothing out-of-scope snuck in.

### Findings

**1. [Medium — usability/correctness] `.well-known/` is blanket-denied.** `static.ts:186` — the dot-segment denylist rejects **every** segment starting with `.`, including `.well-known`. On the CLI's default serving path this breaks ACME HTTP-01 (`/.well-known/acme-challenge/...` → Let's Encrypt cert issuance), `security.txt`, `apple-app-site-association`, and `assetlinks.json`. Verified: `GET /.well-known/acme-challenge` → 404. The plan explicitly flagged `.well-known` as "SHOULD arguably be servable — check what they chose." The choice here is to deny it. Recommend allowlisting a leading `.well-known/` segment (deny all other dotfiles), so the default static server can actually obtain a certificate. Not a security bug, but a real regression for a very common deployment.

**2. [Low — caching semantics] `ETag` is shared across encodings.** `static.ts:237` — the gzip variant and the identity variant carry the **same** `ETag` (confirmed `W/"4-…"` identical). This is *defensible*: it is a **weak** validator (weak ETags denote semantic equivalence, and different content-encodings of the same bytes are semantically equivalent) and `Vary: Accept-Encoding` is set, so shared caches key correctly. Worth a one-line comment acknowledging the weak-ETag choice; no change required. (The plan's "gzip and identity must not share an ETag" targets *strong* ETags; using weak ones satisfies it in spirit.)

**3. [Low] `renderHTML` bypasses conditional/caching and HEAD.** `static.ts:227-233` returns before ETag/Last-Modified/304/HEAD handling, so an `.html` file served through `renderHTML` gets no validators and is fully `readFile`'d even for `HEAD`. Pre-existing behavior shape; acceptable, but note it.

**4. [Nit] `.br` → `application/x-brotli`** (`static.ts:67`) is nonstandard; there is no registered type for a raw `.br` file. `application/octet-stream` (the fallback) would be more honest. Cosmetic.

**5. [Nit] q-value regex rejects `q=.5`.** `static.ts:105` `/^q=(\d+(?:\.\d+)?)$/` requires a leading digit, so `q=.5` (seen in the wild, though technically ill-formed) is ignored and defaults to `q=1`. And `*;q=0` still yields an identity response rather than 406 — pragmatic and safe, just noting the minor deviations.

### Missing from locked scope
None. Every F31/F32 locked item is implemented and covered by a test.

Nice work — the security-critical core (decode-then-dot-check-then-containment, plus the realpath symlink guard) is correct and I couldn't get past it. Addressing finding 1 before it lands on the default CLI path would be my only real ask.

PR #240, opened 2026-07-14
fix: shutdown and close correctness (F25, F27, F34)

pi0x:
Shutdown/close-correctness batch of the v1 stabilization plan.

## F25 — Deno `close(true)` drops its argument
`deno.ts` only ever called the graceful `shutdown()` and wired no `AbortSignal` into `Deno.serve`, so on Deno a double-Ctrl+C printed "Forcibly closing connections…" and then **hung** on an open SSE/WebSocket. Node and Bun already honor the flag.

- Wire an `AbortController` into `Deno.serve` and abort it on `close(true)` — Deno exposes no per-connection close, so aborting the serve signal is the only way to force in-flight connections closed; the graceful `shutdown()` then resolves immediately instead of hanging.

## F27 — plugin registration order + dropped `maxRequestBodySize` in loader mode
- Move `gracefulShutdownPlugin(this)` **after** the `__srvxLoader__` early-return on Deno and Bun (matching Node). Previously both the inner loader-intercepted instance and the outer CLI server attached SIGINT/SIGTERM handlers → two competing countdowns.
- Thread an inline `serve({ maxRequestBodySize })` from the intercepted inner server's options onto the outer CLI server. In loader mode the user's `serve()` is intercepted before it listens, so the limit lived only on the inner server and was silently dropped.

## F34 — graceful shutdown crash + listener leak
- Capture a rejecting `close()` during graceful shutdown (log, then fall through to force-close) instead of letting it surface as an `unhandledRejection` that crashes the process and skips force-close.
- Remove the SIGINT/SIGTERM listeners on every shutdown path **and** on programmatic `close()` (via a thin wrapper), and cancel the deferred force-close timer, so repeated `serve()`/`close()` cycles no longer accumulate listeners or pin each `Server`.

## Tests
New `test/graceful-shutdown.test.ts` — graceful shutdown previously had zero coverage and is auto-disabled under `CI`/`TEST`. Uses `gracefulShutdown: true` to force it on regardless of env:
- **In-process** (plugin driven directly, no real signals): graceful close + listener cleanup; rejecting `close()` falls through to force-close without an unhandledRejection; no listener accumulation across serve/close cycles (real Node adapter).
- **Child process** (real SIGINT): in-flight request completes then exits 0; `gracefulTimeout` honored → force-closes a hung request.
- Added a CLI test asserting the loader threads an inline `maxRequestBodySize` (413 on overflow).

`bun`/`deno` share the same runtime-agnostic `_plugins.ts` path exercised by the in-process tests; F25 was additionally verified with a Deno SSE smoke test (`close(true)` now resolves instead of hanging).

## Results
`pnpm test` green: 983 passed, 35 skipped (pre-existing skips). Lint + typecheck clean.

## Summary by CodeRabbit

* **New Features**
  * Added reliable graceful shutdown handling, including forced closure for hanging requests.
  * Added support for enforcing maximum request body sizes in CLI-served applications.
  * Deno servers can now terminate in-flight connections during forced shutdown.

* **Bug Fixes**
  * Improved cleanup of shutdown signal handlers and prevented unhandled shutdown errors.
  * Fixed shutdown behavior across Bun and Deno loader scenarios.

* **Tests**
  * Added coverage for graceful shutdown, forced termination, signal cleanup, and request size limits.

### DIFF
--- a/src/_plugins.ts
+++ b/src/_plugins.ts
@@ -30,14 +30,33 @@ export const gracefulShutdownPlugin: ServerPlugin = (server) => {
 
   let isClosing = false;
   let isClosed = false;
+  let forceCloseTimer: ReturnType<typeof setTimeout> | undefined;
 
   const w = server.options.silent ? () => {} : process.stderr.write.bind(process.stderr);
 
+  // Remove every listener this plugin registered. Called on all shutdown paths
+  // (graceful, forced, timed-out) and on programmatic `close()` so repeated
+  // serve()/close() cycles don't accumulate listeners or pin each Server.
+  const removeListeners = () => {
+    // Cancel the deferred force-close registration so it can't re-add a listener
+    // after cleanup has already run.
+    if (forceCloseTimer) clearTimeout(forceCloseTimer);
+    globalThis.process.removeListener("SIGINT", shutdown);
+    globalThis.process.removeListener("SIGTERM", shutdown);
+    globalThis.process.removeListener("SIGINT", forceClose);
+  };
+
   const forceClose = async () => {
     if (isClosed) return;
-    w(c.red("\x1b[2K\rForcibly closing connections...\n"));
     isClosed = true;
-    await server.close(true);
+    w(c.red("\x1b[2K\rForcibly closing connections...\n"));
+    try {
+      await nativeClose(true);
+    } catch (error) {
+      w(c.red(`\x1b[2K\rError while force closing connections: ${error}\n`));
+    } finally {
+      removeListeners();
+    }
   };
 
   const shutdown = async () => {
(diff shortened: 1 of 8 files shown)


pi0x (commented):
**Verdict: Approve** (independent review; I did not write this code). Adversarial pass over every shutdown path found no blocking bug. Spec fully met, suite green (983 passed / 35 pre-existing skips, typecheck clean), and `test/graceful-shutdown.test.ts` ran clean 3x with no flakiness.

### Spec checklist
- **F25 (Deno `close(true)` force-closes)** — ✅ `deno.ts:41,120,124,158-160`. An `AbortController` is wired into `Deno.serve({ signal })` and aborted only on `close(true)`; graceful `close()` (no arg) skips the abort and still awaits `server.shutdown()`. Double-Ctrl+C now aborts in-flight SSE/WS and `shutdown()` resolves immediately. `close()` signature widened to `close(closeAll?)`, bringing Deno into line with the frozen `Server.close(closeActiveConnections?)` interface it previously under-implemented.
- **F27 (plugin order)** — ✅ `bun.ts:53-55`, `deno.ts:64-66`. `gracefulShutdownPlugin(this)` moved *after* the `__srvxLoader__` early-return, matching `node.ts:96-103`. Under the CLI only the outer server attaches SIGINT/SIGTERM. The plugin mutates no options (only registers handlers + wraps `close`), so moving it past `wrapFetch`/`serveOptions` loses nothing.
- **F27 (maxRequestBodySize threading)** — ✅ `cli/serve.ts:80-81`. `serverOptions.maxRequestBodySize ?? loaded.srvxServer?.options?.maxRequestBodySize`. Precedence is right: an explicit module-level export wins, else the intercepted inner `serve()` option is threaded onto the outer server. No plugin re-registration, so F27's double-handler issue is not resurrected. CLI 413 test confirms it.
- **F34 (rejecting close)** — ✅ `_plugins.ts:78-80,94-98`. `nativeClose().catch()` captures the rejection into `closeError`; the loop detects it and falls through to `forceClose()` — no unhandledRejection, no skipped force-close. `forceClose` itself wraps `nativeClose(true)` in try/catch/finally.
- **F34 (listener cleanup)** — ✅ `_plugins.ts:40-47,109-111,117-121`. `removeListeners()` runs on graceful completion, force path, and programmatic `close()` (via a thin wrapper capturing the un-wrapped `nativeClose`), and it `clearTimeout`s the deferred force-close registration. Verified: the "second Ctrl+C forces" flow is intact — during the graceful window none of the cleanup paths fire, so the 100ms `forceClose` once-listener is registered and stays available; the timer is only cancelled when graceful close finishes in <100ms (when force isn't needed). No path leaves neither graceful nor force completing.
- **Tests** — ✅ Env auto-disable is bypassed correctly via explicit `gracefulShutdown: true` (fixture + in-process mocks). In-process tests exercise the real plugin: graceful+cleanup, rejecting-close fall-through (unhandledRejection listener wired; also fails via awaited throw on the old code), and no-accumulation across 3 real serve/close cycles. Child-process tests use real SIGINT with `getRandomPort` + `waitForPort` (200 retries) and SIGKILL cleanup in `finally`.

### Findings (all minor / non-blocking)
- **[low] `test/graceful-shutdown.test.ts:11` `grabShutdownListener()`** grabs the *last* SIGINT listener, assuming it's the plugin's `shutdown`. Fragile if any other code (vitest, another plugin) registers a SIGINT listener after the plugin in the same worker. Works today; consider capturing the count delta or the handler ref instead.
- **[low] child-process fixture** (`fixtures/graceful-server.ts`, `fixtures/cli-max-body/server.ts`) is a raw `.ts` run via `process.execPath` with no `--import`, relying on Node native type-stripping (fine on the repo's Node 24, would break on Node <22.6). Consistent with existing CLI tests; note for the support matrix.
- **[cosmetic] `_plugins.ts`** double-Ctrl+C prints both "Forcibly closing connections…" and then "Server closed successfully." (the graceful loop observes the force-closed server and reports success). Harmless.
- **[note] API shape** the `close` wrapper converts a prototype method into an own arrow property — but only when the plugin is active (post early-return guard), so the default `TEST`/`CI` env and the v1-api-freeze checks still see the original method. Fine.

### Expected merge conflicts with #230 (both touch `_plugins.ts`, `bun.ts`, `deno.ts`)
- **`_plugins.ts:27`** — direct conflict: #230 changes `!config?.gracefulTimeout` → `config?.gracefulTimeout == null` (the F9 fix #240 correctly leaves alone), while #240 adds code a few lines below at 30-33. Trivial: keep #230's `== null` line + #240's new lines.
- **`bun.ts` / `deno.ts` constructor** — semantic conflict: #230's hunks carry `gracefulShutdownPlugin(this)` in its *original* (pre-early-return) position, whereas #240 *moves* it after the loader early-return (F27). Whoever lands second must preserve #240's moved position while keeping #230's constructor changes (F26 listen-error wrapping, `errorPlugin` import+call). Also a one-line import conflict (`import { gracefulShutdownPlugin }` vs `{ errorPlugin, gracefulShutdownPlugin }`).
- **`deno.ts` close()** — #230 has a hunk near the close method and #240 rewrites `close()` (adds the abort). Small overlap; resolvable.

Recommend #230 and #240 land with an explicit rebase/conflict-resolution step rather than trusting auto-merge. Scope boundaries respected: #240 does not touch the F9 `??` fix, F26 ready-rejects, or cloudflare/bunny/service-worker.

PR #241, opened 2026-07-14
fix: web core T3 polish (plugin contract, tracing, log, ready)

pi0x:
Web-core **T3 polish** batch of the v1 stabilization plan. Excludes PR #230 items (F51/F52/F10/F11 `@experimental`/D6). Prose docs left to the docs agent (JSDoc only here).

## Changes

1. **Plugin extension contract** (`_middleware.ts`) — `wrapFetch` now reads `server.options.middleware` **live on every request** instead of snapshotting at construction. Previously the behavior was inconsistent: if the chain was empty at construction, later `server.options.middleware.push(...)` (the mechanism the docs teach) was silently ignored; if non-empty, pushes took effect. Now push-after-construction always works on both paths. Documented in JSDoc. Compatible with D6 (sync-only plugins → all plugin middleware is in place before the first request).

2. **`deno` runtime type** (`types.ts`) — added `server?: Deno.HttpServer` to the request runtime `deno` context; `deno.ts` already sets it and the docs document it.

3. **Tracing** (`tracing.ts`) — the frozen middleware `index` went stale and internal middleware were untraced, because `errorPlugin`/`trustProxyPlugin` add their middleware *after* user plugins run. Fixed by installing a dispatch-time per-middleware interceptor (via an internal symbol read by `wrapFetch`) instead of snapshotting/rewrapping the array at plugin time. The interceptor sees the **final** live chain: every middleware is traced and `index` reflects its real position. Event shape is unchanged (`srvx/tracing` is experimental per D5 regardless).

4. **`log()`** (`log.ts`) — `LogOptions` gains `sink?: (line) => void` (default `console.log`) and `colors?: boolean` (default: auto — TTY and not `NO_COLOR`, `FORCE_COLOR` forces on), evaluated at factory time. Added `test/log.test.ts`.

5. **`ready()` contract** (`types.ts`) — **Chosen fix: align the JSDoc with reality** (smallest honest fix). Making `ready()` block until `serve()` is invoked would hang forever under `manual: true` and touch every adapter. The JSDoc now states it resolves once listening has started, and resolves immediately if listening was never initiated — pointing callers at `serve()`'s returned promise for a hard guarantee. No behavior change; does not touch F26 (owned by #230).

6. **`package.json` exports** — added `"./package.json": "./package.json"`. Left the single-string subpaths as-is (TS resolves the `.d.mts` beside each `.mjs`); adding explicit `{types,default}` objects is mechanical but pure churn, so noted rather than done.

7. **Build hygiene** (`build.config.mjs`) — an `end` hook strips the dead `declare module "./types.ts"` augmentation from the emitted `dist/mtls.d.mts` (it only survived via `skipLibCheck`). The `srvx` augmentation consumers need is retained. Verified with `pnpm build`.

## Tests

`pnpm test` (lint + typecheck + vitest + coverage): **984 passed, 35 skipped (pre-existing), no type errors**. `pnpm build` clean; confirmed the stripped `dist/mtls.d.mts` no longer contains the dead block. New tests: `test/middleware.test.ts`, `test/log.test.ts`, and a tracing case asserting the internal (`error`) middleware is traced at index 0 with the user middleware shifted to index 1.

### DIFF
--- a/build.config.mjs
+++ b/build.config.mjs
@@ -1,7 +1,41 @@
 import { defineBuildConfig } from "obuild/config";
-import { rm } from "node:fs/promises";
+import { readFile, rm, writeFile } from "node:fs/promises";
 import { join } from "node:path";
 import pkg from "./package.json" with { type: "json" };
+
+/**
+ * Remove a `declare module "<spec>" { ... }` augmentation block from emitted `.d.mts`.
+ *
+ * `src/mtls.ts` augments both `"srvx"` (for consumers) and `"./types.ts"` (so srvx's
+ * own type-check resolves `request.tls` through the relative source module). The latter
+ * is meaningless in the published bundle — `./types.ts` does not exist there — and only
+ * survives into `dist/mtls.d.mts` because `skipLibCheck` hides it. Strip it at build time.
+ */
+function stripDeclareModule(code, spec) {
+  const marker = `declare module "${spec}" {`;
+  const start = code.indexOf(marker);
+  if (start === -1) {
+    return code;
+  }
+  let depth = 0;
+  let end = start + marker.length - 1; // index of the opening brace
+  for (; end < code.length; end++) {
+    const ch = code[end];
+    if (ch === "{") {
+      depth++;
+    } else if (ch === "}") {
+      depth--;
+      if (depth === 0) {
+        end++;
+        break;
+      }
+    }
+  }
+  if (code[end] === "\n") {
+    end++;
+  }
+  return code.slice(0, start) + code.slice(end);
+}
 
 export default defineBuildConfig({
   entries: [
@@ -47,6 +81,11 @@ export default defineBuildConfig({
   hooks: {
     async end(ctx) {
       await rm(join(ctx.pkgDir, "dist/types.mjs"));
+
+      // Strip the dead `declare module "./types.ts"` augmentation (see helper above).
+      const mtlsDts = join(ctx.pkgDir, "dist/mtls.d.mts");
+      const contents = await readFile(mtlsDts, "utf8");
+      await writeFile(mtlsDts, stripDeclareModule(contents, "./types.ts"));
     },
   },
 });
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
   "sideEffects": false,
   "types": "./dist/types.d.mts",
   "exports": {
+    "./package.json": "./package.json",
     "./deno": "./dist/adapters/deno.mjs",
     "./bun": "./dist/adapters/bun.mjs",
     "./node": "./dist/adapters/node.mjs",
--- a/src/_middleware.ts
+++ b/src/_middleware.ts
@@ -1,23 +1,56 @@
 import type { Server, ServerRequest, ServerHandler, ServerMiddleware } from "./types.ts";
 
+/**
+ * Internal per-middleware interceptor hook.
+ *
+ * When set on a server, {@link wrapFetch} invokes it for every middleware in the
+ * live chain instead of calling the middleware directly, passing the handler and
+ * its current index. Used by the experimental `srvx/tracing` plugin so that
+ * tracing reflects the *final* middleware chain (including middleware that
+ * internal plugins such as `trustProxy`/`error` add after user plugins run).
+ *
+ * @internal
+ */
+export const kMiddlewareInterceptor: unique symbol = Symbol.for("srvx.middlewareInterceptor");
+
+/** @internal */
+export type MiddlewareInterceptor = (
+  handler: ServerMiddleware,
+  index: number,
+  request: ServerRequest,
+  next: () => Response | Promise<Response>,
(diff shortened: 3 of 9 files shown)


pi0x (commented):
**Verdict: Approve with comments.** The diff is correct standalone — 984 tests pass, `build`/`typecheck`/`lint` clean, the mtls `.d.mts` hook produces correct output, and I verified the extension contract, tracing rework, and type additions against the live code. No hard hot-path regression that blocks. The findings below are low-severity or merge-coordination; the one item that genuinely needs action before merge is reconciliation with #230, which rewrites the same `wrapFetch`.

### Per-item checklist
1. Plugin extension contract (live read) — ✅ correct. Both empty and non-empty initial paths now pick up post-construction `push`. JSDoc specifies the contract; `test/middleware.test.ts` covers both paths.
2. `types.ts` deno `server?: Deno.HttpServer` — ✅ matches `deno.ts:75,112`. Accurate.
3. Tracing rework (dispatch-time interceptor) — ✅. Same events/order for user middleware; internal middleware (error/trustProxy) now traced with correct final-chain indices (intended per plan; `test/tracing.test.ts` asserts it). No per-request leak: interceptor set once, no array mutation (strictly better than the old splice).
4. `log.ts` sink + colors — ✅ with two minor notes below.
5. `ready()` JSDoc-only — ✅ doc-only, does not touch F26 behavior.
6. `package.json` `"./package.json"` export — ✅.
7. Build `end` hook strips dead `declare module "./types.ts"` — ✅ verified against `pnpm build` output; `declare module "srvx"` augmentation preserved.

### Findings by severity

**MEDIUM — merge coordination with #230 (`src/_middleware.ts`)**
PR #230 rewrites the *same* `wrapFetch` to add `request.context ??= {}` (F51 lazy-init). This PR's rewrite omits it. Whichever merges second hits a conflict, and if resolved carelessly F51 silently regresses — the new `wrapFetch` (`_middleware.ts:37`) has no `request.context` init. Fold it into the new dispatch path when rebasing: `(request) => { request.context ??= {}; return callMiddleware(...) }`. #230 and #241 also both edit the `ready()` JSDoc and the `deno` region of `types.ts` → textual conflicts there too. Not a bug in this diff in isolation, but must be reconciled before merge.

**LOW/MEDIUM — build hook is not string/comment-aware (`build.config.mjs:22-38`)**
The brace matcher scans raw text. A stray or unbalanced `{`/`}` inside a JSDoc comment or string literal within the augmented block would truncate the removal and corrupt `dist/mtls.d.mts`, shipping silently (no assertion). Currently safe (the block's braces are balanced; `{@link ...}` are balanced) and it safely no-ops if the block disappears upstream (`indexOf` → -1). Suggest a cheap guard: after the loop, assert `depth === 0` (a matching close was found) and that the result still contains `declare module "srvx"` before `writeFile`; otherwise throw so a future obuild/JSDoc shift fails the build loudly instead of shipping a broken `.d.mts`.

**LOW — mid-dispatch front-mutation skew (`src/_middleware.ts:46-51`)**
`callMiddleware` reads `server.options.middleware` live at *each* recursion level, so `unshift`/`splice` on the array while a request is in flight causes index skew (skip/double-run; I reproduced unbounded recursion in a pathological self-unshifting middleware). `push` — the mechanism the docs teach and the JSDoc guarantees — is safe (appended entries are reached in order). This hazard already existed for the non-empty case in the old code (same live-array-by-reference), so it is **not a regression**, and plugins are sync/in-place before the first request. Worth one sentence in the contract JSDoc: the guarantee is about `push` before/between requests, and front-mutation during in-flight dispatch is unsupported.

**LOW — `FORCE_COLOR=0` treated as force-on (`src/log.ts:31`)**
`if (env.FORCE_COLOR)` treats any non-empty value including `"0"` as force-colors-on, whereas the common chalk/supports-color convention treats `FORCE_COLOR=0` as force-*off*. NO_COLOR handling is correct: empty string is falsy → not disabled, matching the no-color.org "present and non-empty" rule.

**INFO — fast-path dispatch cost (`src/_middleware.ts:37,46-48`)**
The common `serve({ fetch })` case (no error/trustProxy/middleware → empty array) no longer short-circuits to `fetchHandler`; every request goes through the wrapper closure + two property reads + a length check. I micro-measured the added cost at single-digit nanoseconds/request — immaterial against real request work, and the correctness of the live contract justifies it. Not a blocker; flagging only because the task called out the hot path. The `kMiddlewareInterceptor` symbol lookup only happens once middleware is present, and is a cheap undefined-property miss when tracing is off.

**INFO — CLI color inconsistency (`src/log.ts` / `src/cli/serve.ts`)**
Moving color detection to factory time is correct and does not break the CLI (`serve.ts:91` calls `log()` in the main process where `stdout.isTTY` is accurate). But the CLI banner lines (`serve.ts:160-168`, `main.ts:65`) still use `cli/_utils` colors unconditionally, so under a pipe or `NO_COLOR` the request-log lines drop color while the banners keep it. Cosmetic.

PR #242, opened 2026-07-14
refactor: improve edge adapters and tests

pi0x:
Edge-adapter batch of the v1 stabilization plan.

### Cloudflare — F29 (locked: keep the global `fetch` listener, fix the bugs)
- **No double-registration**: `serve()` retains its listener; repeated calls (or `manual: true` then `serve()`) no longer stack a second listener, which previously caused a double `respondWith()` error.
- **`close()` removes the listener** it registered.
- **Threads `env`** through the service-worker listener instead of a hardcoded `{}`. Cloudflare exposes bindings as globals in service-worker syntax, so `env` is only meaningfully available via **module-worker** syntax — documented in the class JSDoc.
- **`request.ip` is now `configurable: true`** (matching bun/deno) so `trustProxy` can override it; still populated from `cf-connecting-ip`.

### Bunny
- Guard the `waitUntil` closure with `typeof Bunny !== "undefined"` so it no longer throws a `ReferenceError` when the adapter runs outside the Bunny runtime (e.g. `manual: true`).

### Service worker
- `close()` unregisters **only** the registration `serve()` created, instead of every service worker on the origin.

### F46 — adapter coverage
- `test/generic.test.ts` — in-process tests (middleware, plugins, `error`, `waitUntil`/`close`).
- `test/bunny.test.ts` — mocked `Bunny` global (registration, single-registration, `waitUntil` delegation, `ip`, and the no-runtime `ReferenceError` regression).
- `test/cloudflare.test.ts` — best-effort **in-process** test with mocked service-worker globals + direct module-worker `fetch(request, env, context)` calls, covering all three F29 behaviors and the configurable `ip`.
  - Note: the plan's preferred `@cloudflare/vitest-pool-workers` smoke job needs its own worker runtime + wrangler + a separate vitest project, which conflicts with the existing single-config Node suite, so the in-process fallback was used (as the plan allows).

### F9 — error paths (one unhandled-throw test per runtime)
- `test/_error-fixture.ts` — a server with **no** `error` option (so `errorPlugin` no-ops and throws propagate to the runtime).
- Deno & Bun (`test/_error-tests.ts`, wired into `deno.test.ts` / `bun.test.ts` — zero extra CI wiring): both answer **500** and keep serving.
- Node (`test/node-error-paths.test.ts`, in-process): documents Node's current behavior — an unhandled throw escapes as an `uncaughtException` and leaves the connection hanging (an unguarded server would crash). Making Node answer 500 like Deno/Bun belongs to the Node adapter scope, so it is **documented, not fixed** here.

## Verification
- `pnpm lint` clean, `pnpm typecheck` clean (after `pnpm build`), full `vitest run`: **1002 passed / 35 skipped**.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Bunny runtime compatibility when accessing optional runtime APIs.
  - Prevented duplicate Cloudflare service-worker request handlers during repeated server starts.
  - Ensured Cloudflare shutdown removes only the handler registered by the server.
  - Improved proxy IP configuration flexibility for Cloudflare requests.
  - Service-worker shutdown now unregisters only the registration created by the server.
  - Improved resilience when request handlers throw synchronous or asynchronous errors, helping servers remain available.

### DIFF
--- a/src/adapters/bunny.ts
+++ b/src/adapters/bunny.ts
@@ -44,7 +44,8 @@ class BunnyServer implements Server {
 
     const fetchHandler = wrapFetch(this);
 
-    const waitUntil = (this.waitUntil = (p: Promise<unknown>) => (globalThis as any).Bunny?.unstable?.waitUntil?.(p));
+    const waitUntil = (this.waitUntil = (p: Promise<unknown>) =>
+      (globalThis as any).Bunny?.unstable?.waitUntil?.(p));
 
     this.fetch = (request: Request) => {
       Object.defineProperties(request, {
--- a/src/adapters/bunny.ts
+++ b/src/adapters/bunny.ts
@@ -44,7 +44,15 @@ class BunnyServer implements Server {
 
     const fetchHandler = wrapFetch(this);
 
-    const waitUntil = (this.waitUntil = (p: Promise<unknown>) => Bunny.unstable?.waitUntil?.(p));
+    const waitUntil = (this.waitUntil = (p: Promise<unknown>) => {
+      // `Bunny` is a bare global that only exists inside the Bunny runtime.
+      // Referencing it directly (even with optional chaining) throws a
+      // `ReferenceError` when the adapter runs elsewhere (e.g. `manual: true`
+      // in tests or another runtime), so guard with a `typeof` check first.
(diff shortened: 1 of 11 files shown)


pi0x (commented):
.

PR #243, opened 2026-07-14
fix(node): Node adapter correctness fixes (statusText, empty body, HEAD streaming, send errors, sync bridge)

pi0x:
## Summary

A batch of correctness fixes for the Node adapter, aligning its behavior with native `fetch`/`Response`/`Request` and with Deno/Bun. Each fix has a regression test in `test/node-adapters.test.ts`.

## Fixes

- **`req.text()` / `req.json()` reject instead of throwing.** Reading a body stream that's already locked or disturbed (e.g. something else consumed `req.body` directly) now returns a rejected promise, matching native `fetch`, instead of throwing synchronously.

- **`statusText` defaults to `""`, not Node's `"OK"`.** `NodeResponse` no longer looks up Node's `http.STATUS_CODES` table for a default reason phrase. It now matches native `Response`, Bun, and Deno, which all default to an empty string.

- **Empty-string response bodies get correct headers.** `new NodeResponse("")` previously skipped the `content-type: text/plain` and `content-length: 0` headers because `""` is falsy. Both are now set correctly, matching native `Response("")`.

- **HEAD responses no longer pump the body stream.** A streaming body attached to a HEAD response is now cancelled immediately instead of read to completion — previously an unbounded stream (e.g. SSE) would pump forever on a HEAD request. Matches Deno/Bun.

- **Send errors are now logged.** If sending a response fails synchronously (e.g. an invalid header), the error is now logged via `console.error` (unless the server was created with `silent: true`) instead of being swallowed. The client still just gets a bare 500.

- **Synthetic `IncomingMessage` now reports HTTP version and raw headers.** The Node-compat bridge used for legacy `(req, res)` handlers previously left `httpVersion`/`httpVersionMajor`/`httpVersionMinor` at their defaults and didn't populate `rawHeaders`, breaking things like morgan's `:http-version` token. These are now filled in, including proper `set-cookie` array handling. Because reporting HTTP/1.1 makes `ServerResponse` chunk-encode by default, the bridge's response is explicitly kept un-chunked so the raw captured body isn't corrupted by chunk framing.

- **Documented the `loader.ts` handler-detection heuristic.** No behavior change — added comments explaining how a bare `export default function` is disambiguated between a web `fetch` handler and a legacy Node handler (by parameter count), and the known edge cases where that guess is wrong.

- **Clarified `_request` / `_url` JSDoc.** `_request`'s doc incorrectly said it held "Node.js native instance of request" — it actually holds the web-standard `Request`. Fixed the wording.

Not included: a fix for `_request`/`headers` snapshot identity was attempted and reverted (`1754b02`) — it trades early-reference liveness for a frozen headers snapshot and needs its own discussion. Tracked separately in #245. Also out of scope: no `upgrade` listener support (documented limitation, not a regression from this PR).

## Testing

`pnpm test` (lint + typecheck + vitest --coverage) passes.

### DIFF
--- a/src/adapters/_node/request.ts
+++ b/src/adapters/_node/request.ts
@@ -133,23 +133,10 @@ export const NodeRequest: {
     }
 
     get headers(): Headers {
-      // Return the srvx-side Headers object whenever it exists, even after the
-      // native Request has materialized. Materializing `_request` always reads
-      // `this.headers` (so `#headers` is populated) but the native constructor
-      // copies the entries into its own Headers instance. Switching the getter
-      // to `this.#request.headers` there would leave a reference taken earlier
-      // (`const h = req.headers; req._request; h.set(...)`) pointing at a now-
-      // detached object whose mutations are invisible. Keeping `#headers`
-      // canonical keeps those references live.
-      if (this.#headers) {
-        return this.#headers;
-      }
       if (this.#request) {
-        // Unreachable in practice: materializing `_request` reads `this.headers`
-        // first, so `#headers` is always set. Kept as defense in depth.
         return this.#request.headers;
       }
-      return (this.#headers = new NodeRequestHeaders(this.#req));
+      return (this.#headers ||= new NodeRequestHeaders(this.#req));
     }
 
     get _abortController() {
@@ -299,11 +286,7 @@ export const NodeRequest: {
           // @ts-expect-error Undici specific
           duplex: body ? "half" : undefined,
         });
-        // Keep `#headers` so `get headers()` returns the same object identity
-        // before and after materialization (see the note there). The native
-        // Request holds its own snapshot copy: mutations made after this point
-        // show up in `req.headers` but not in `#request.headers` — and thus not
-        // in `clone()` or `formData()`, which delegate to the native Request.
+        this.#headers = undefined;
         this.#bodyStream = undefined;
       }
 
--- a/src/adapters/_node/request.ts
+++ b/src/adapters/_node/request.ts
@@ -133,10 +133,21 @@ export const NodeRequest: {
     }
 
     get headers(): Headers {
+      // Return the srvx-side Headers object whenever it exists, even after the
+      // native Request has materialized. Materializing `_request` always reads
+      // `this.headers` (so `#headers` is populated) but the native constructor
+      // copies the entries into its own Headers instance. Switching the getter
+      // to `this.#request.headers` there would leave a reference taken earlier
+      // (`const h = req.headers; req._request; h.set(...)`) pointing at a now-
+      // detached object whose mutations are invisible. Keeping `#headers`
+      // canonical keeps those references live.
+      if (this.#headers) {
+        return this.#headers;
+      }
       if (this.#request) {
         return this.#request.headers;
       }
-      return (this.#headers ||= new NodeRequestHeaders(this.#req));
+      return (this.#headers = new NodeRequestHeaders(this.#req));
     }
 
     get _abortController() {
@@ -232,7 +243,14 @@ export const NodeRequest: {
       }
       this.#bodyUsed = true;
       if (this.#bodyStream !== undefined) {
-        return new Response(this.#bodyStream).text();
+        // `new Response(stream)` throws *synchronously* if the stream is already
+        // locked/disturbed (e.g. a consumer took `req.body` and read it
+        // directly). Surface that as a rejected promise, matching native fetch.
+        try {
+          return new Response(this.#bodyStream).text();
+        } catch (error) {
+          return Promise.reject(error);
+        }
       }
       return this.#readBuffered().then((buf) => buf.toString());
     }
@@ -250,7 +268,12 @@ export const NodeRequest: {
       }
       this.#bodyUsed = true;
       if (this.#bodyStream !== undefined) {
-        return new Response(this.#bodyStream).json();
+        // See text(): a locked/disturbed stream must reject, not throw.
+        try {
+          return new Response(this.#bodyStream).json();
+        } catch (error) {
+          return Promise.reject(error);
+        }
       }
       // Parse in a single continuation (readBody -> parse) instead of going
       // through text() — one less promise + microtask hop per body read.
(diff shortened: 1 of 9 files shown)


pi0x (commented):
## Review: PR #243 — fix(node): T3 correctness batch

**Verdict: Approve (non-blocking).** I checked out the branch, read every touched file in full, and ran the suites. `pnpm vitest run test/node-adapters.test.ts test/node.test.ts` → 159 passed / 6 skipped. Full `pnpm test` (after `pnpm build` so the `srvx` self-imports resolve) → **985 passed / 35 skipped, no type errors, lint clean**. Diff is scoped strictly to the Node adapter + loader + types + tests; no scope creep, no `upgrade` listener added, F17/F18 in `response.ts`/`node.ts` left untouched (owned by #230). No correctness bugs found — the changes are careful and the reasoning in the comments holds up. A few informational notes and one expected merge conflict below.

### Per-item checklist (1–11)
1. **F53 `types.ts`** ✅ `_request` JSDoc now correctly describes the undici web `Request` (not `IncomingMessage`, points at `runtime.node.req`); both `_request`/`_url` marked supported/frozen (D7).
2. **Loader `--import` message** ✅ Verified `loader.ts:166` already reads `srvx --import jiti/register` on main; this PR does not touch it. Claim correct.
3. **Loader arity heuristic** ✅ Comment/JSDoc added only; `mod.default.length < 2` condition byte-identical. No behavior change.
4. **`handleSendError` diagnostics** ✅ `console.error`s unless `silent`; logged **once** (only from the detached catch — `sendNodeResponse` still rejects, streaming errors stay `.catch(()=>{})`). `silent` is threaded via `this.options.silent`; the changed function is `@internal` (`sendNodeResponse`, the public re-export, is unchanged) → **no public signature change**. Client still gets a bare 500, no detail leak (test asserts empty body).
5. **HEAD stream cancel** ✅ `stream.cancel().catch(()=>{})` is error-guarded; headers are already written by the caller before `streamBody`, so content-type/length still emitted; non-stream HEAD bodies fall to Node's own `_hasBody=false` suppression. Correct.
6. **`statusText` → `""`** ✅ Defaults to spec empty phrase; Node still fills a reason phrase into the status line on the wire (`_storeHeader` uses `STATUS_CODES` when `statusMessage` is falsy), so wire output stays valid. Test asserts at the web-Response layer, which is the right layer.
7. **Empty-string body** ✅ `!= null` (not truthy) now emits `text/plain; charset=UTF-8` + `content-length: 0`. See note A on the exact string.
8. **`text()`/`json()` on locked/disturbed stream** ✅ `new Response(stream)` synchronous throw wrapped → `Promise.reject`. Test drives it via a directly-disturbed `req.body` and asserts a `TypeError` rejection.
9. **GET/HEAD null-body** ✅ Verified `#hasBody()` already on main (#228); PR only adds the locked-stream regression test.
10. **Materialization identity (headers)** ✅ for the requirement in the plan — see note B for the residual divergence.
11. **Synthetic IncomingMessage + chunked fix** ✅ — scrutinized hard, see below.

**On item 11 (the one that needed the hardest look):** the `useChunkedEncodingByDefault = false` in `web/response.ts` is **correct and necessary, not incidental**. Reporting HTTP/1.1 flips Node's `ServerResponse` default: previously the synthetic `httpVersionMajor` was `null` (`null < 1` → true), so Node's constructor took the `<1.1` branch and set `useChunkedEncodingByDefault = false`; now that branch is skipped and it inherits `true` from `OutgoingMessage`. The bridge's `WebRequestSocket._write` captures the post-header bytes **verbatim** (no de-chunking) into `_webResBody`, so without this line every non-`content-length` streamed bridge response would be chunk-framed and corrupted. The bridge socket is a `Duplex`, never a real wire, so disabling the default has no on-wire effect. `rawHeaders` is built as flat `[name, value, …]` preserving case, with `set-cookie` expanded per-entry and stored as an array on `headers` (matching Node) — good.

### Findings

**A. [informational] Empty-string content-type differs from native undici.** `_node/response.ts:130` emits `text/plain; charset=UTF-8` (space, `UTF-8`) whereas native `Response("")` gives `text/plain;charset=UTF-8`. This is the adapter's long-standing convention for **all** string bodies, not new to this change — the PR's own test documents the mismatch. Consistent within the adapter; flagging only so it's a conscious call.

**B. [low] `req.headers` and `req._request.headers` diverge after materialization.** `_node/request.ts` now keeps `#headers` (the `NodeRequestHeaders`) canonical across `_request` materialization — which correctly satisfies the plan's "a reference taken before materialization stays live." But the native `Request` built in the `_request` getter copies header entries into its **own** `Headers` at construction. So after `void req._request`, mutating `req.headers` (or the pre-materialization ref) is **not** visible on `req._request.headers`, and vice-versa. Since `_request` is now documented supported API, a consumer that materializes, mutates `req.headers`, then hands `req._request` to undici/`fetch` will silently lose the mutation. The added test only exercises the `req.headers`/old-ref direction, not the `_request` direction. Not a regression (pre-PR the pre-materialization ref detached instead), and arguably acceptable, but the boundary is worth a doc line or a follow-up.

**C. [informational] Body identity across materialization is punted.** The `_request` comment addresses headers only; `#bodyStream = undefined` is retained so post-materialization `body` comes from the native request. No regression (body reads funnel consistently through one source), but the plan flagged body identity too and the PR neither covers nor explicitly calls it out. A one-line "body identity intentionally deferred" note would close the loop.

**D. [informational] `rawHeaders` cannot fully reconstruct wire form.** `web/incoming.ts` builds `rawHeaders` from `req.headers.entries()`, which comma-joins duplicate non-`set-cookie` headers, so repeated headers collapse to one entry and original wire order isn't preserved. The comment acknowledges "best-effort." Acceptable for the morgan/`:http-version` use case.

**E. [merge — expected conflict with #230]** Both PRs edit `_node/response.ts` in the same region. #243 **deletes** `const STATUS_CODES = …` and rewrites `statusText` to drop it; #230 **keeps** `STATUS_CODES` (its `statusText` still reads it) and inserts `NULL_BODY_STATUS` + a null-body constructor throw immediately after that line. This will conflict textually and semantically — resolution must keep **#243's** `statusText`/`STATUS_CODES` removal **and** **#230's** null-body constructor throw. `types.ts` (#230 removes netlify/stormkit/vercel at ~347–353; #243 edits `_request`/`_url` JSDoc at ~361–386) and `node.ts` (#230 at lines 6/54/140/194; #243 at 86–92) are non-adjacent and should auto-merge.

Nice work — the hot-path reasoning (chunked-default restoration, once-only error logging, canonical-headers identity) is sound and well-commented.

pi0 (approved):

PR #266, opened 2026-07-17
perf(log): batch stdout writes and cache the timestamp

pi0x:
The `log()` middleware called `console.log` once per request and rebuilt an `Intl`-backed timestamp every time. This buffers lines and flushes once per event-loop turn, caches the timestamp, and stops emitting colors in production.

## Changes

- **Batch writes.** Lines buffer and flush once per event-loop turn (`setImmediate`, or `queueMicrotask` where it's unavailable) instead of one `console.log` per request. Under real concurrency many requests completing in the same turn share a single write. A `drain`-based backpressure path keeps a slow stdout from growing the buffer without bound.
- **Cache the timestamp.** `toLocaleTimeString()` is Intl-backed and cost more than the rest of the line combined, yet its second-resolution output changes only once a second — so it's built once per second, not once per request.
- **No colors in production.** ANSI escapes are noise once logs reach a file or a collector. Piping already stripped them, but a container started with a tty (`docker -t`, systemd) still got them. Resolved per `log()` call (the CLI sets `NODE_ENV` after this module is imported); `FORCE_COLOR` overrides.

Output is byte-identical to before. A line still buffered at process exit is flushed synchronously from an `exit` hook.

## Benchmark

(removed in https://github.com/h3js/srvx/pull/266/commits/78875cee7b5ee49e9d11baf164e7b460571247cb to simplify diff)

`pnpm bench:log` drives a real server with `oha` at 64 connections and compares three servers — **no logger**, the **previous** `console.log`-per-request logger (vendored verbatim from the pre-change commit, [`_before.mjs`](test/bench-log/_before.mjs)), and the **current** `log()` — across the sinks logs actually go to. Batching only shows under real concurrency, so an in-process loop can't measure it.

```
Intel i7-10700K · Node v24.18.0 · oha 1.14.0 · 64 conns · median of 3×3s
```

RPS for each logger, with its overhead versus `no logger` in parentheses:

| sink | no logger | old logger | new logger | diff |
| --- | ---: | ---: | ---: | ---: |
| `terminal` | 20,713 | 13,594 (-34.4%) | 18,260 (-11.8%) | **+34.3%** |
| `pipe` | 20,297 | 15,900 (-21.7%) | 18,559 (-8.6%) | **+16.7%** |
| `devnull` | 19,788 | 15,564 (-21.3%) | 19,084 (-3.6%) | **+22.6%** |

- **`terminal`** (a real TTY) is the worst case for the old logger — a synchronous colored `console.log` per request blocks on the terminal's rendering. It gave up 34% of throughput; the current logger gives up 12%.
- **`pipe`** (logs streamed to a collector) is the production shape: **+16.7%**.
- **`devnull`** isolates the CPU/formatting cost — the timestamp cache — from any I/O: **+22.6%**.

The `terminal` figure depends on the terminal emulator's rendering speed, so it's less reproducible than `pipe`/`devnull`; run-to-run noise is a few percent throughout.

## Notes

- **Colors resolve per `log()` call, not at module scope** — the CLI assigns `NODE_ENV` *after* importing this module, so a module-level check reads it too early.
- **No runtime-specific code:** Node, Deno and Bun all go through `process.stdout`; workers fall back to `console.log`.
- Adds the `log()` tests **F46** asks for (the module had zero coverage).
- Partially addresses the `log.ts:4` item — colors are gated and `console.log` is no longer hard-coded; **`LogOptions` still has no sink option**, so that item stays open.

`pnpm test` green, lint and types clean.


package.json
```diff
@@ -41,6 +41,7 @@
     }
   },
   "scripts": {
+    "bench:log": "node test/bench-log/_run.mjs",
```
pi0:
```suggestion
```

pi0 (approved):

PR #274, opened 2026-07-17
feat(log): flush on default-handled SIGTERM/SIGINT and add `batch` option

pi0x:
Follows up on the crash-safety concern raised on #266 ("missed logs prior to a crash"). I stress-tested the batched logger against real process deaths (child processes on Node 22/24, Deno, Bun) and closed the gaps that are actually closable.

## What the stress test showed

Scenarios where every buffered line already arrived before this PR: clean exit, same-tick `process.exit()`, uncaught exception, unhandled rejection (Node runs `exit` hooks on its default crash path), and SIGTERM/SIGINT with srvx graceful shutdown active.

Scenarios that lost the final event-loop turn of lines:

- **SIGHUP/SIGTERM/SIGINT with *default* handling** — no listener anywhere, so the runtime terminates without emitting `exit`. Happens when graceful shutdown is off (explicit `false`, the CI/TEST env carve-out, or `log()` used standalone with a non-srvx server). **Fixable — fixed here.**
- **SIGKILL / OOM killer / V8 heap OOM (SIGABRT)** — no userland code runs. Not fixable by hooks, but the pre-#266 unbatched logger *did* survive these for already-logged lines (each write lands in the kernel pipe buffer synchronously). **Addressed via an opt-out.**

## Changes

- **Signal flush hooks.** `hookExit()` now also registers a SIGHUP/SIGTERM/SIGINT listener (signal-exit style): if it turns out to be the *only* listener at delivery time, it flushes synchronously, removes itself and re-raises — the process still dies by the signal with exit semantics untouched (verified: parent observes `signal: SIGTERM`, not an exit code). When any other listener exists (e.g. srvx graceful shutdown), it defers entirely and the regular flush path delivers the lines. Registration uses `prependListener` so a `process.once(sig, cleanup)` registered earlier is still counted at delivery time — `once` wrappers self-remove before dispatch, and an appended handler would miscount them as absent and re-raise mid-cleanup. The listeners are unref'd on all three runtimes, so an idle process still exits naturally (covered by a test).
- **`batch: false` option.** Writes each line in the request's own turn instead of coalescing per event-loop turn — narrows a hard kill's loss window to the write in flight, matching the pre-#266 logger (verified: lines survive a same-tick SIGKILL on Node, Deno and Bun). It funnels through the shared buffer + drain logic, so it respects backpressure and can never overtake lines a batched logger is still holding; `flush()` gained a mid-drain guard so a stale scheduled flush can't feed a full stream when both modes are mixed.

## Verification

All child-process scenarios re-run after the change — buffered lines delivered in every runnable path:

| scenario | before | after |
| --- | --- | --- |
| default-handled SIGHUP / SIGTERM / SIGINT | ❌ lost | ✅ flushed, dies by signal |
| SIGTERM with graceful shutdown / other listener (incl. `once`) | ✅ | ✅ (hook defers) |
| same-tick `process.exit()`, uncaught exception, natural end | ✅ | ✅ |
| same-tick SIGKILL, `batch: false` | ❌ lost | ✅ in kernel buffer |
| same-tick SIGKILL, batched | ❌ | ❌ (≤ 1 turn, inherent) |

Tests spawn `test/fixtures/log-crash.mjs` and assert on what actually reaches the pipe plus how the process died — crash-flush behavior only shows in a process that really terminates. Suite also gained `batch: false` write/backpressure/mixed-mode cases and listener cleanup so repeated `log.ts` imports don't trip the max-listeners warning.

An independent clean-context review validated the design and surfaced the `once` race, the missing SIGHUP, the mid-drain guard and a doc overclaim — all addressed in the second commit (details in the comment below).

`pnpm test` green (1215 passed), lint and types clean.

## Summary by CodeRabbit

- **New Features**
  - Added a `batch` option for request logging.
  - Logging is batched by default; setting `batch: false` writes each request log immediately.
  - Added crash-safe flushing for buffered logs during normal exits, signals, and uncaught exceptions.

- **Documentation**
  - Expanded middleware logging guidance with batching examples and shutdown behavior.

- **Bug Fixes**
  - Improved reliability of log output during process termination and stdout backpressure.

### DIFF
--- a/docs/1.guide/4.middleware.md
+++ b/docs/1.guide/4.middleware.md
@@ -76,6 +76,12 @@ serve({
 
 The duration is measured around `next()`, so place `log()` first in the array for it to cover the whole chain. The [CLI](/guide/cli) enables this middleware automatically.
 
+**`log()` options:**
+
+- `batch`: Coalesce lines and write once per event-loop turn (default: `true`). Set to `false` to hand each line to stdout synchronously — slower under concurrency, but a line is never buffered in the process when it dies.
+
+Buffered lines are flushed on process exit, including uncaught exceptions and `SIGTERM`/`SIGINT`. Only a hard kill (`SIGKILL`, the OOM killer) can drop the lines of the final event-loop turn — set `batch: false` where that window matters more than throughput.
+
 ### Static files
 
 `serveStatic()` from `srvx/static` serves files from a directory, with `index.html` resolution, `.html` extension fallback (`/about` → `about.html`), common MIME types, gzip/Brotli compression, and path-traversal protection.
--- a/docs/1.guide/4.middleware.md
+++ b/docs/1.guide/4.middleware.md
@@ -78,9 +78,16 @@ The duration is measured around `next()`, so place `log()` first in the array fo
 
 **`log()` options:**
 
-- `batch`: Coalesce lines and write once per event-loop turn (default: `true`). Set to `false` to hand each line to stdout as its request completes — slower under concurrency, but nothing waits for a flush turn when the process is killed.
+- `batch`: Coalesce lines and write once per event-loop turn (default: `true`). Set to `false` to hand each line to stdout as its request completes — slower under concurrency, but a line never waits for a flush turn.
 
-Buffered lines are flushed on process exit, including uncaught exceptions and `SIGHUP`/`SIGTERM`/`SIGINT`. A hard kill (`SIGKILL`, the OOM killer) can still drop the lines of the final event-loop turn — `batch: false` narrows that window to the write in flight.
+```ts
+serve({
+  middleware: [log({ batch: false })],
+  fetch: () => new Response("👋 Hello there."),
+});
+```
+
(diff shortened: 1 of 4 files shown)


pi0x:
Ran an independent clean-context review over the implementation; it confirmed the core mechanism but found four issues, fixed in b645cd4:

1. **`process.once` race in the sole-listener check.** A `once` cleanup handler registered *before* the logger self-removes ahead of dispatch, so the logger's appended handler saw `listenerCount === 1` and re-raised mid-cleanup. The handlers are now registered with `prependListener`, so they run before any `once` wrapper disappears from the count. Covered by the new `once-cleanup` fixture scenario.
2. **SIGHUP was missing** — it default-terminates without `exit` exactly like SIGTERM/SIGINT (terminal close is its common delivery), and the docs claimed only hard kills could drop lines. Added to the hook and to the fixture tests.
3. **Mixed batched + unbatched flush invariant.** A flush scheduled by a batched line could fire after an unbatched write had started a drain wait, feeding the full stream and stacking a duplicate `drain` listener. `flush()` now early-returns while draining (which also let `writeNow` drop its own guard). Pinned by the new mid-drain test.
4. **Doc overclaim.** `batch: false` lines still buffer during a drain wait, and a piped stdout write is asynchronous — so the honest guarantee is "narrows a hard kill's loss window to the write in flight", not "never buffered". Wording adjusted in the JSDoc and guide.

Full suite green after the changes (1215 passed), scenarios re-verified on Node, Deno and Bun.

pi0 (approved):

pi0 (approved):

Issue #284, opened 2026-07-29
bug: shared adapter type chunk breaks tsc for Node-only consumers without skipLibCheck


cany748:
### Environment

srvx 0.12.4, TypeScript 7.0.2, @types/node 26, Node 24

### Reproduction

`tsconfig.json`: `{"module": "NodeNext", "target": "esnext", "lib": ["esnext"], "strict": true}`

```ts
import {staticMiddleware} from "srvx/static";

export const mw = staticMiddleware({dir: "./public"});
```

`npx tsc --noEmit` → 13 errors, all inside srvx's own `.d.mts`.

### Describe the bug

`dist/_chunks/types.d.mts` is one shared chunk pulled in by every entrypoint, including the Node-only `srvx/node` and `srvx/static`. It imports `aws-lambda`, `bun` and `@cloudflare/workers-types`, and references the `Deno` namespace and `FetchEvent`. All of those are `devDependencies` of srvx, so consumers never receive them — a Node-only project cannot type-check without `skipLibCheck: true`, even though it targets none of those runtimes.

`skipLibCheck` is repo-wide, so working around one dependency also stops checking every other dependency's types.

Separately, `dist/adapters/node.d.mts:43` uses the global `BodyInit`, which `@types/node` does not provide and which needs `"lib": ["dom"]`. Importing it from `undici-types` would avoid that.

### Additional context

Possible fixes, in order of preference:

1. Split the type chunk per adapter, so Node-only entrypoints stop referencing Bun/Deno/Workers/Lambda types. This looks like the root fix — the coupling comes from how `.d.mts` is bundled, not from a real dependency.
2. Declare `@types/aws-lambda`, `@types/bun`, `@cloudflare/workers-types` and `@types/deno` as `peerDependencies` with `peerDependenciesMeta: {"optional": true}`.
3. Ship fallback ambient declarations when those packages are absent.

Happy to open a PR for (1) if you agree with the direction.

### Logs

```
dist/_chunks/types.d.mts(4,22):   error TS2307: Cannot find module 'aws-lambda' or its corresponding type declarations.
dist/_chunks/types.d.mts(6,22):   error TS2307: Cannot find module 'bun' or its corresponding type declarations.
dist/_chunks/types.d.mts(7,21):   error TS2307: Cannot find module '@cloudflare/workers-types' or its corresponding type declarations.
dist/_chunks/types.d.mts(196,10): error TS2503: Cannot find namespace 'Deno'.
[...]

PR #285, opened 2026-07-29
fix(types): declare runtime type packages as optional peer dependencies

pi0x:
Refs #284 (option 2 of the three listed there).

`dist/_chunks/types.d.mts` is shared by every entrypoint and references `bun`, `aws-lambda`, `@cloudflare/workers-types`, the `Deno` namespace and `FetchEvent`. All of those were only `devDependencies`, so consumers had no declared way to know which packages the published types expect.

This declares them — plus `@types/node` — as `peerDependencies` with `peerDependenciesMeta.optional: true`, and adds a TypeScript section to the getting-started guide listing the package per runtime. Ranges are `*` because these are ambient type packages: any version supplies the namespaces, and pinning would only produce spurious peer warnings.

Nothing is installed automatically — npm does not auto-install optional peers (verified) — so this is purely declarative for existing users.

### What this does and does not fix

Measured on the issue's reproduction (`srvx@0.12.4`, `module: NodeNext`, no `skipLibCheck`), from 12 errors in the shared chunk:

- **Fixed** once the packages are installed: `Cannot find module 'aws-lambda' | 'bun' | '@cloudflare/workers-types'`.
- **Not fixed:** `Cannot find namespace 'Deno'` (5x) and `Cannot find name 'FetchEvent'` need `@types/deno` / `@types/serviceworker` to be globally included, and those two cannot be loaded in the same program: `@types/serviceworker` redeclares globals (`Blob`, `Request`, `GPU*`, …) that `@types/deno` declares too (13 errors), and that `@types/bun` declares too (49). Substituting TypeScript's own `lib: ["webworker"]` for `FetchEvent` collides with `@types/bun` all the same (54). Pairwise they are fine — `@types/node` with any single one of them type checks at 0 errors — so the shared chunk needing all of them at once is the actual blocker.
- **Not fixed:** `Cannot find module 'cloudflare:workers'` (2x). `@cloudflare/workers-types` does declare that module, but only as an ambient declaration in a global script file, so importing the package as a module does not bring it in.
- `dist/adapters/node.d.mts:43` `BodyInit` is untouched — `@types/node` genuinely does not provide that global.

So this is a documentation/declaration improvement, not a complete fix: a Node-only consumer still cannot type check srvx's declarations with `skipLibCheck: false`. #286 is the root fix — it splits the type chunk so each entry references only its own runtime, which is what makes these packages installable one at a time.

## Summary by CodeRabbit

* **Documentation**
  * Added TypeScript guidance explaining that `srvx` provides a shared set of types across supported runtimes.
  * Included recommendations for pairing runtime type packages and compiler settings (including `skipLibCheck`) to avoid missing-type and global redeclaration issues.
  * Clarified `@types/deno` DOM type reference and suggested `lib` configuration.

* **Improvements**
  * Declared `@cloudflare/workers-types` and `@types/*` packages as optional peer dependencies for supported runtimes (e.g., Node.js, Deno, Bun, AWS Lambda, Service Worker).

### DIFF
--- a/docs/1.guide/01.index.md
+++ b/docs/1.guide/01.index.md
@@ -83,6 +83,25 @@ bun run server.mjs
 
 ::
 
+## TypeScript
+
+srvx ships one shared set of type definitions covering every supported runtime. The runtime types they reference come from ambient type packages, declared as optional `peerDependencies` so that nothing is installed unless you ask for it:
+
+| Runtime            | Types package               |
+| ------------------ | --------------------------- |
+| Node.js            | `@types/node`               |
+| Deno               | `@types/deno`               |
+| Bun                | `@types/bun`                |
+| Cloudflare Workers | `@cloudflare/workers-types` |
+| AWS Lambda         | `@types/aws-lambda`         |
+| Service Worker     | `@types/serviceworker`      |
+
+> [!NOTE]
+> `@types/deno` references DOM types, so pair it with `lib: ["esnext", "dom"]`.
+
+> [!NOTE]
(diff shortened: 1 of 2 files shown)


pi0:
closing in favor of #287

PR #286, opened 2026-07-29
fix(types): split runtime types per entry

### LINKED ISSUE #284: bug: shared adapter type chunk breaks tsc for Node-only consumers without skipLibCheck

pi0x:
Closes #284 — option 1 from the issue: make `skipLibCheck` unnecessary instead of only declaring what to install.

### The problem

`dist/_chunks/types.d.mts` was one chunk shared by every entry, because `ServerOptions`, `Server` and `ServerRuntimeContext` name Bun/Deno/Workers/Lambda/Service Worker types directly. Anything reaching `ServerRequest` — `srvx/static`, `srvx/log`, `srvx/mtls` — dragged all of them in.

### The change

Runtime declarations move to `src/_types/<runtime>.mts`. Each registers itself by augmenting a `RuntimeTypes` registry declared in the core module:

```ts
// src/_types/deno.mts
declare module "./core.mjs" {
  interface RuntimeTypes {
    "deno.options": Deno.ServeOptions;
    "deno.context": { info: Deno.ServeHandlerInfo<Deno.NetAddr> };
  }
}
```

```ts
// src/_types/core.mts — no runtime references
deno?: RuntimeType<"deno.options">;
```

Public shape is unchanged: every member keeps its name, and is precisely typed as soon as that runtime's declarations are loaded (importing `srvx`, or `srvx/deno` anywhere in the project). When they are not, the field resolves to `Record<string, unknown>` rather than disappearing, so `serve({ deno: {…} })` still compiles in a cross-runtime server and `request.runtime?.deno?.info` reads as `unknown` instead of erroring.

Also here:

- **`srvx/types`** — new runtime-free types entry (`dist/_types/core.d.mts`), which is what middleware authors want.
- **`@types/serviceworker` is no longer needed at all.** `srvx/service-worker` now resolves `FetchEvent` through a `typeof globalThis extends { FetchEvent: … }` probe, falling back to the small `FetchEventLike` shape srvx actually uses. Consumers who do have the package (or `lib: ["webworker"]`) still get the real `FetchEvent` — verified: a handler annotated `event: FetchEvent` reading `event.clientId` still assigns to `ServiceWorkerHandler`. It is dropped from the peer deps, and kept as a devDependency for srvx's own adapter sources.
- **`BodyInit`** now derives from the ambient `Response` (`NonNullable<ConstructorParameters<typeof globalThis.Response>[0]>`), fixing the last error from the issue without needing `lib: ["dom"]` or `undici-types`.
- **`srvx/mtls`** augmented `"srvx"` plus a `"./types.ts"` mirror that dangles in the published output; both are replaced by one augmentation of the core module, so opting into `srvx/mtls` no longer pulls in the universal entry.
- The runtime type packages are declared as optional `peerDependencies`, and the guide gains a TypeScript section mapping each runtime to its package.

Two build details, both commented in `build.config.mjs`: `preserveEntrySignatures: "allow-extension"` keeps `_types/*` as real entries instead of facades over a shared chunk, and the `end` hook rewrites the augmentation target from the relative specifier (which only resolves where the sources live) to `<pkg>/types` (which resolves from any chunk, and stays correct for `srvx-nightly`).

### Verified

Consumer project, `module: NodeNext`, `skipLibCheck: false`, only `@types/node` installed — errors reported by `tsc`:

| Import | before | after |
| --- | --- | --- |
| `srvx/static` (the issue's repro) | 12 | **0** |
| `srvx/node` | 13 | **0** |
| `srvx/service-worker` | 13 | **0** |
| `srvx/log` + `srvx/body-limit` + `srvx/mtls` | 12 | **0** |
| `srvx/types` | n/a | **0** |
| `srvx` (universal) | 12 | 11 |

The universal entry covers every runtime by design, so with only `@types/node` it still reports the runtimes it cannot resolve. The difference is that those are now installable: with the five packages present (`types: node,deno,bun`, `lib: ["esnext","dom"]`) it drops to **2** errors, both `cloudflare:workers`. `@cloudflare/workers-types` only declares that module ambiently, so it resolves when the package is in tsconfig `types`, which conflicts with `@types/deno`; declaring the module yourself (3 lines, now documented in the guide) takes even the universal entry to **0** errors without `skipLibCheck`. Dropping `@types/serviceworker` is what unblocked this — it was the package that collided with `@types/bun` (49 errors) and `@types/deno` (13); the remaining four coexist cleanly.

With all packages installed, precision is unchanged: `server.node?.server`, `server.bun?.server.port`, `request.runtime?.node?.req`, `runtime?.cloudflare?.context`, `runtime?.awsLambda?.context` and `BunFetchHandler` all still resolve, and an unknown key in `node: { … }` still errors.

- `pnpm lint`, `pnpm typecheck`, `pnpm vitest run` (31 files, 1298 tests) green, including the existing `test/types.test-d.ts` assertions for `runtime.cloudflare.env` and the mtls `request.tls` augmentation.
- Every emitted `.mjs` is byte identical to a build of `main` — this is a types-only change.
- `test/dist-types.test.ts` walks the declaration closure of each entry and fails if a runtime type leaks into a runtime-agnostic one (verified against an injected leak).

### Follow-up worth considering

`exports["."]` could gain per-condition `types`, so `import { serve } from "srvx"` resolves to the adapter of the runtime you target and the universal entry stops being the fat one. It is safe now that missing runtimes degrade to loose types instead of errors, but it changes what bare `srvx` means per resolver, so I left it out of this PR.


pi0:
closing in favor of #287

PR #303, opened 2026-09-07
fix(node): handle promises independently of the global constructor

### LINKED ISSUE #302: Node adapter returns empty 500s after Zone.js replaces global Promise
The Node adapter returns an empty HTTP 500 for native async-handler promises after Zone.js replaces `globalThis.Promise`. The first request can succeed when it lazily loads Zone.js; subsequent requests fail before the handler's async result is sent.

## Minimal reproduction

Tested on Linux x64, Node 24.15.0, `srvx@1.0.3`, and `zone.js@0.16.1`. No Angular, Nitro, or Vite installation is needed.

```sh
npm init -y
npm install srvx@1.0.3 zone.js@0.16.1
```

Save as `repro.mjs` and run `node repro.mjs`:

```js
import { serve } from 'srvx/node';

const request = globalThis.fetch;
const server = serve({
  port: 0,
  hostname: '127.0.0.1',
  silent: true,
  async fetch() {
    await import('zone.js/node');
    return new Response('ok');
  },
});
await server.ready();

try {
  for (let i = 0; i < 3; i++) {
    const response = await request(server.url);
    console.log(i + 1, response.status, JSON.stringify(await response.text()));
  }
} finally {
  await server.close();
}
```

Actual:

```text
1 200 "ok"
2 500 ""
3 500 ""
```

Expected: all three requests return `200 "ok"`.

## Cause and use case

[The current Node listener](https://github.com/h3js/srvx/blob/d4c95fe20ec1342dfe34f575a491c0241b65b911/src/adapters/node.ts#L102) tests `res instanceof Promise`. Native async functions return intrinsic native promises; after Zone.js installs its Promise constructor, those results fail the check and are passed to the response sender as though they were Responses. The 1.0.3 `toNodeHandler` adapter also contains an `instanceof Promise` dispatch check worth auditing.

This surfaced while qualifying Analog's Angular integration with Nitro `3.0.260903-beta`. A real Zone.js Angular SSR page rendered on the first direct Vite request, then subsequent requests returned empty 500s. Nitro's development runner exports an async fetch entry. Both `self` and `node-worker` development configurations exhibited failures. Plain renderer fixtures without Zone.js passed, and a separately built production app passed successive dynamic SSR requests, so this report is deliberately scoped to the demonstrated promise-dispatch failure rather than claiming every Angular deployment is affected.

Promise values from another realm or a replaced global constructor are also worth covering in the same fix.

## Suggested AI implementation prompt

> Fix the Node adapter's response dispatch so valid asynchronous handler results are recognized even when globalThis.Promise is replaced, as with Zone.js, or the promise comes from another realm. Keep the synchronous Response fast path and preserve existing rejection handling so send errors are not answered twice. Audit the analogous toNodeHandler bridge. Avoid modifying the user's global Promise or buffering responses. Add a regression that reproduces the first-request success followed by repeated 500s above, and focused cases for native promises, an alternate Promise constructor/realm, synchronous Responses, rejected handlers, and streamed responses. Verify server shutdown and run the repository's required checks. Keep this a scoped compatibility fix to async-result detection.

benpsnyder:
## Problem and change

When Zone.js replaces global Promise, native async functions still return intrinsic native promises. The Node adapter's `instanceof Promise` check then treats a promise as a Response and returns an empty 500 on subsequent requests.

Closes [Node responses after Zone.js replaces Promise — h3js/srvx#302](https://github.com/h3js/srvx/issues/302).

Discriminate the existing Response/Promise union through a **callable** `then` property in both Node adapter paths. Apply the same change to custom error middleware, using the rejection arm of `then`, so native/cross-realm rejections still reach the configured error handler. Synchronous responses retain the direct send path; no new dependency or global patch is introduced.

## Sample and use case

```js
const server = serve({
  async fetch() {
    await import('zone.js/node');
    return new Response('ok');
  },
});
```

Repeated requests should return `200 "ok"`. This affects Angular/Analog SSR development under Nitro, where loading Zone.js changes the global constructor after server startup. The linked issue contains the complete standalone reproduction.

The new tests replace global Promise with a subclass to reproduce the relevant constructor mismatch without adding Zone.js as a dependency, restore it after each test, and also exercise promises created in a separate VM realm.

## Validation

Node 24.15.0, repository-pinned pnpm 11.11.0:

- `pnpm install --frozen-lockfile`
- `pnpm build`
- `env -u NO_COLOR -u FORCE_COLOR pnpm test`: lint, formatting, typecheck and coverage suite passed; **1,478 tests passed, 39 skipped**, no type errors.
- New focused suite: 10 passed, including Responses with a non-callable `then` property. Against unmodified upstream source, the same suite produced 5 failures and 2 unhandled-rejection errors.

Tests cover three successive requests through `serve` and `toNodeHandler`, native and cross-realm promises, synchronous responses, and rejections with/without a custom error handler. The initial full run inherited conflicting terminal-color variables and failed logger assertions, plus one TLS test; the clean-environment full rerun passed all tests. Build precedes typecheck to resolve package self-imports through `dist`.

## Suggested AI follow-up prompt

> Audit async response boundaries for assumptions about a mutable global Promise constructor. Preserve synchronous response fast paths and existing rejection behavior. Reproduce native async and cross-realm results with isolated tests, including configured error handlers, without globally normalizing or buffering responses.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of asynchronous responses across Node-based integrations, including cross-environment promises and thenable values.
  - Prevented invalid or empty handler results from causing unexpected runtime errors; these now produce a standard 500 response.
  - Preserved correct error handling for rejected asynchronous responses, including custom error responses when configured.
  - Responses with non-callable `then` properties are now handled as regular responses.

### DIFF
--- a/src/adapters/_node/adapter.ts
+++ b/src/adapters/_node/adapter.ts
@@ -1,3 +1,4 @@
+import { isAsyncResponse } from "../../_middleware.ts";
 import type {
   FetchHandler,
   NodeHttpHandler,
@@ -34,7 +35,7 @@ export function toNodeHandler(
       trustProxy: options?.trustProxy,
     });
     const res = handler(request);
-    return "then" in res
+    return isAsyncResponse(res)
       ? res.then((resolvedRes) => send(nodeRes, resolvedRes))
       : send(nodeRes, res);
   }
--- a/src/adapters/_node/adapter.ts
+++ b/src/adapters/_node/adapter.ts
@@ -34,7 +34,7 @@ export function toNodeHandler(
       trustProxy: options?.trustProxy,
     });
     const res = handler(request);
-    return res instanceof Promise
+    return "then" in res
       ? res.then((resolvedRes) => send(nodeRes, resolvedRes))
       : send(nodeRes, res);
   }
--- a/src/adapters/_node/adapter.ts
+++ b/src/adapters/_node/adapter.ts
@@ -1,4 +1,3 @@
-import { isAsyncResponse } from "../../_middleware.ts";
 import type {
   FetchHandler,
   NodeHttpHandler,
@@ -35,9 +34,9 @@ export function toNodeHandler(
       trustProxy: options?.trustProxy,
     });
     const res = handler(request);
-    return isAsyncResponse(res)
+    return "then" in res && typeof res.then === "function"
       ? res.then((resolvedRes) => send(nodeRes, resolvedRes))
-      : send(nodeRes, res);
+      : send(nodeRes, res as Response);
   }
 
   (convertedNodeHandler as AdapterMeta).__fetchHandler = handler;
--- a/src/adapters/_node/adapter.ts
+++ b/src/adapters/_node/adapter.ts
@@ -34,8 +34,8 @@ export function toNodeHandler(
       trustProxy: options?.trustProxy,
     });
     const res = handler(request);
-    return "then" in res && typeof res.then === "function"
-      ? res.then((resolvedRes) => send(nodeRes, resolvedRes))
+    return typeof (res as Promise<Response>)?.then === "function"
+      ? (res as Promise<Response>).then((resolvedRes) => send(nodeRes, resolvedRes))
       : send(nodeRes, res as Response);
   }
 
--- a/src/adapters/node.ts
+++ b/src/adapters/node.ts
@@ -99,7 +99,7 @@ class NodeServer implements Server {
       }
       // node:http ignores the listener's return value — use the detached
       // variant to skip the per-response end-tracking Promise.
-      return res instanceof Promise
+      return "then" in res
         ? res.then(
             (resolvedRes) => sendNodeResponseDetached(nodeRes, resolvedRes, this.options.silent),
             // Rejection handler (not `.catch`) so send failures, which
--- a/src/adapters/node.ts
+++ b/src/adapters/node.ts
@@ -8,7 +8,7 @@ import {
   resolvePortAndHost,
   createWaitUntil,
 } from "../_utils.ts";
-import { wrapFetch, isAsyncResponse } from "../_middleware.ts";
+import { wrapFetch } from "../_middleware.ts";
 import { errorPlugin, gracefulShutdownPlugin } from "../_plugins.ts";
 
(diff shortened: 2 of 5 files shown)


src/adapters/node.ts
```diff
@@ -99,7 +99,7 @@ class NodeServer implements Server {
       }
       // node:http ignores the listener's return value — use the detached
       // variant to skip the per-response end-tracking Promise.
-      return res instanceof Promise
+      return "then" in res
```
pi0: we need a safer check like `typeof <>.then === "function"` otherwise it will catch any `.then` property

src/adapters/_node/adapter.ts
```diff
@@ -34,7 +35,7 @@ export function toNodeHandler(
       trustProxy: options?.trustProxy,
     });
     const res = handler(request);
-    return res instanceof Promise
+    return isAsyncResponse(res)
```
pi0: It is only used in 3 places lets inline logic without util (PS: I am not sure but it might affect js engine optimizations / IC entries)

src/adapters/node.ts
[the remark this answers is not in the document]
benpsnyder: Fixed in 9e3d4f5. All three paths require a callable `then`, with regression coverage for `then: undefined`. Full suite: 1,478 tests pass.

benpsnyder: Fixed in 9e3d4f5 with an inline callable-then check at all three sites. The non-callable-then regressions and full suite pass.

src/adapters/_node/adapter.ts
benpsnyder: Inlined at all three call sites and removed the helper in 9e3d4f5. Full suite: 1,478 tests pass.

benpsnyder:
Review feedback is addressed and all three threads are resolved; the latest automated review found no new actionable comments. Local lint, build, and all 1,478 tests pass. GitHub Actions needs maintainer approval, and the Vercel check needs team authorization.

pi0 (approved):
Thnx!

pi0 (approved):

